PR #8, opened 2023-04-29
added dependencies and custom exceptions

HenriqueAJNB:
# Proposed changes

- Added `isort` to `pre-commit` and as a dev dependency
- Added `openai` to dependencies.
- Apply `isort` to the entire project (purely style).
- Added a custom exception to API Key Not Found error.


gventuri:
Great job @HenriqueAJNB. I'm just wondering if we want to add `openai` as a dependency. As of now, it's pretty much the only llm that is able to produce significant results, but do we really want to have it as a dependency? What do you think?

gventuri:
@HenriqueAJNB let's add it the `openai` package for now, then decide what to do in the future in case the package gets bigger. I've fixed some of the issues reported by `pylint`, and as part of it we are now using `openai` as a dependency.

HenriqueAJNB:
@gventuri, I just found that it should produce some bugs if the dependencies are not well defined. 

If it's just a test one, we can add it separately into group test, for example, on poetry dependencies.

gventuri:
@HenriqueAJNB I'd say let's leave it as it is, so let's keep `openai` installed, considering it's the only working LLM as of now. Then in the future we might think about conditional packages!

Would you mind having a look at the merge conflicts? For example, I've added some other custom exceptions all over the codebase and some files are now conflicting.

HenriqueAJNB:
Maybe it's better to push it again if conflicts are hard to solve. I can handle this in the best way @gventuri

gventuri:
Sounds great! Then feel free to go for it @HenriqueAJNB!

PR #10, opened 2023-04-29
Add CI

FillerFree:


FillerFree:
There are lot of lint errors added pylint. Will add black to automatically fix lint errors in separate PR

Issue #15, opened 2023-05-01
Include show code feature


johnson-52197:
Hi @gventuri,

I have included a feature that let's the user view the code that was used to generate the answer. A parameter named show_code can be included in the run function which will create a new cell under the prompt and pastes the code in it. Let me know if I can integrate it. 

Thank you
<img width="952" alt="Screenshot 2023-05-01 at 6 34 22 AM" src="https://user-images.githubusercontent.com/58318959/235386591-ff53bc8a-644c-43e2-81b6-2cbbcaf0f9cb.png">

gventuri:
Hey @johnson-52197, definitely a super nice-to-have feature, would love to have it integrated. I've assigned it to you!

P.s. we are launching the discord server, in particular for the contributors, but more in general for everyone interested into playing around with PandasAI!

If you are interested:
https://discord.gg/kF7FqH2FwS

PR #19, opened 2023-05-02
Added show_code feature

johnson-52197:
#15 

If show_code is set to True, it will create a new cell and paste the code that was used to generate the answer.


gventuri:
Some tests are failing: could you check it out?

Also, maybe we should find a way to make this feature only available if it's in a notebook.
Check it out: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook

What do you think?

johnson-52197:
Sure @gventuri I'll check the failing cases. 

Thank you for the resource, I'll include the ipython notebook checks as well.

johnson-52197:
All the unit test cases have passed successfully when running the application locally.

johnson-52197:
Yet to include the notebook check.

gventuri:
@johnson-52197 I'd say in that case we could `self.log()`, so it will only print it `if self._verbose == True:`. What do you think?

johnson-52197:
That makes more sense @gventuri. I'll modify it.

johnson-52197:
@gventuri included all the changes. Please review it.

pandasai/__init__.py
```diff
@@ -172,6 +193,11 @@ def run_code(
             return result
         except:
             return captured_output
+        
+    def create_new_cell(self, contents: str):
```
gventuri, line 197: Maybe we can move it to a dedicate helper (like notebook or anything like that?

In this way we keep the code cleaner.

pandasai/__init__.py
```diff
@@ -59,6 +62,18 @@ def __init__(
         self._is_conversational_answer = conversational
         self._verbose = verbose
         self._enforce_privacy = enforce_privacy
+        self._is_notebook = self.in_notebook()
+
+    def in_notebook(self):
```
gventuri, line 67: Maybe we can move it to a dedicate helper (like notebook or anything like that?

In this way we keep the code cleaner.

gventuri (changes requested):
Almost there @johnson-52197!

johnson-52197:
Sure @gventuri. 

I could think of two approaches.
1. Creating a folder named helpers inside pandasai and include notebook.py (we can add more helpers in the future)
2. a single python file named notebook.py inside pandasai.

Which one would you prefer?

pandasai/__init__.py
johnson-52197: > Sure @gventuri.
> 
> I could think of two approaches.
> 
> 1. Creating a folder named helpers inside pandasai and include notebook.py (we can add more helpers in the future)
> 2. a single python file named notebook.py inside pandasai.
> 
> Which one would you prefer?

gventuri:
Great job @johnson-52197! I like the 1 better! It scales very well, let's go for it! 🚀

johnson-52197:
@gventuri done. Please review.

gventuri (approved):

Issue #32, opened 2023-05-04
error requesting data in `json` format


avelino:
**traceback:**
```
Traceback (most recent call last):
  File "/Users/avelino/projects/buser/openapi/recommendation-next-travel/use-pandasai.py", line 43, in <module>
    print(pandas_ai.run(
          ^^^^^^^^^^^^^^
  File "/Users/avelino/projects/buser/openapi/recommendation-next-travel/.venv/lib/python3.11/site-packages/pandasai/__init__.py", line 120, in run
    answer = self.run_code(code, data_frame, False)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/avelino/projects/buser/openapi/recommendation-next-travel/.venv/lib/python3.11/site-packages/pandasai/__init__.py", line 166, in run_code
    exec(code)
  File "<string>", line 9, in <module>
NameError: name 'json' is not defined
```

**source:**

```py
data = {...}
df = pd.DataFrame(data)
llm = OpenAI()
pandas_ai = PandasAI(llm)

print(pandas_ai.run(
    df,
    prompt="""
    suggest when the next trip will be, destination and seat type,
    in JSON format: {"destination": "", "seat-type": ""}
    """))
```

## proposed solution

how **I solved it locally**, but I believe it is not the best way

add `import json` in `__init__.py`

gventuri:
@avelino turns out the error was a different one. In some cases, OpenAI was returning an output like this:

```
result = {"happiness": 0.5, "gdp": 0.8}
print(result)
```

There was a conflict between the "print" listener and trying to print a variable that had been declared!

PR #33, opened 2023-05-04
interpret json when llm returns json code

avelino:
fixed: #32

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -1,7 +1,6 @@
 """ PandasAI is a wrapper around a LLM to make dataframes convesational """
 import io
 import sys
-import json
 import pandas as pd
 from .llm.base import LLM
 from .helpers.notebook import Notebook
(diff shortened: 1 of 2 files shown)


gventuri:
Hey @avelino, thanks a lot for the great job!

I'd just ask you if you can also provide a test for this specific use case. Not that we have done a great job so far, but as the project grows, we should find a way to improve our coverage. What do you think?

avelino:
@gventuri I believe the test is validating

avelino:
@gventuri did you remove the `import json`?

gventuri:
@avelino yes, as the response as of now is a json string, no need to import json library from what I've seen.

The error that I could find was about a different issue:
```
result = "{ param: '', param2: '' }"
print(result)
```

In fact, we intercept `print` calls and try to return `result`, but it wasn't defined because of a wrong implementation and so it was raising error.

However, if you want we can add support for json and in that case convert it to a dict.

Let me know what you think, I'd like to be more specific if my explanation wasn't clear.

PR #34, opened 2023-05-04
Small refactorings

HenriqueAJNB:
At this PR I've suggested some small changes for refactoring.


gventuri:
All good, great job @HenriqueAJNB! 🎉

I just noticed there's an additional error during the linting (the one about Exception on line 180). Could you fix it?

pandasai/__init__.py
```diff
@@ -176,9 +176,8 @@ def run_code(
             return eval(last_line[6:-1])
         # Evaluate last line and return its value or the captured output
         try:
-            result = eval(last_line)
-            return result
-        except:
+            return eval(last_line)
+        except Exception:
```
gventuri, line 180: The linter complains about using a generic exception!

gventuri (changes requested):

pandasai/__init__.py
HenriqueAJNB: Since here should be a generic exception, I'll add a pylint disable.

PR #36, opened 2023-05-04
isort added to pre-commit (CI/CD)

HenriqueAJNB:
@gventuri, I've preferred to open another PR just for adding `isort`. Better than solving those conflicts from PR #8 which has a mix of changes.

This one:
- Adds `isort` as a dev dependency
- Apply `isort` with `black` profile in entire project
- Adds `isort` to `pre-commit`
- Suggests a reminder at `README.md` for installing `pre-commit` before contributing to follow our standards.


HenriqueAJNB:
The best way to solve this is to add `pytest` also to `pre-commit`. This avoids commits that failure the tests.

And I'll try to run the tests and check what is going on. It seems to be missing import of `re` module.

HenriqueAJNB:
The best way to solve this is to add `pytest` also to `pre-commit`. This avoids commits that failure the tests.

And I'll try to run the tests and check what is going on. It seems to be missing import of `re` module.

HenriqueAJNB:
Ready to merge!

HenriqueAJNB:
I've also added `pytest` to run on every commit to avoid commits in PR that will failure the unit tests.

gventuri:
Great job @HenriqueAJNB!! 🎉

PR #55, opened 2023-05-05
Anonymized data frame head

alfonsograziano:
The main goal of this PR is to anonymize a few data (like email, phone numbers and credit cards) and shuffle the non-sensitive data. 

The reason behind it is that we're sending a data frame head to a remote LLM (like OpenAI GPT) and we want to be sure that the info cannot be used by anyone in an unwanted way.

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -87,7 +87,7 @@ Rewrite the answer to the question in a conversational way.
         prompt: str,
         is_conversational_answer: bool = None,
         show_code: bool = False,
-        anonymize_df: bool = False,
+        anonymize_df: bool = True,
     ) -> str:
         """Run the LLM with the given prompt"""
         self.log(f"Running PandasAI with {self._llm.type} LLM...")
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -8,6 +8,7 @@ import pandas as pd
 from .exceptions import LLMNotFoundError
 from .helpers.notebook import Notebook
 from .llm.base import LLM
+from .helpers.anonymizer import anonymize_dataframe_head
 
 
 class PandasAI:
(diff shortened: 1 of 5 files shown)


pandasai/helpers/anonymizer.py
```diff
@@ -0,0 +1,130 @@
...
+    name_length = random.randint(6, 12)
+    domain = random.choice(domains)
+    letters = string.ascii_lowercase + string.digits + "-_"
+    username = "".join(random.choice(letters) for i in range(name_length))
+    email = username + "@" + domain
+    return email
+
+
+def generate_random_phone_number():
+    """
```
gventuri, line 65: Noticed here we have a problem: sometimes the script detects a "fake number" (check out the example of the GDP). The result is we convert it the gpd to a phone number like this: `+60 4702356135`, which can be misleading and cause hallucinations to the LLM.

What if we add the country prefix (i.e. +60) only if it was originally present, otherwise we just generate a random number without prefix?

So:
- if the field is, for example, 2411255037952
   - we generate a random number, say 5263101311
- if the field is, for example, +39 11255037952
   - we generate a random number, say +39 5263101311

pandasai/__init__.py
```diff
@@ -86,21 +87,27 @@ def run(
         prompt: str,
         is_conversational_answer: bool = None,
         show_code: bool = False,
+        anonymize_df: bool = False,
```
gventuri, line 90: We can also use it as a default value IMO, so I'd set to True by default! What do you think?

gventuri (changes requested):
Super, great job @alfonsograziano!!

Left you a couple of minor improvements.
Plus, could you write a couple of tests? We need to increase the coverage, as this library is growing and being used as a dependency in other projects, we need to be very careful!

Thank you so much and great job!

gventuri:
P.s. also check out the linter (run `pylint` locally).

pandasai/helpers/anonymizer.py
alfonsograziano: Addressed, thanks. Great catch :D

gventuri (approved):

PR #64, opened 2023-05-06
add requirements

jianboy:
* add requirements

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+pandas
+dotenv
+openai
+requests


gventuri:
Hey @jianboy, thanks a lot for the PR. However, we've decided to use `poetry` as a dependency manager, so we don't need to add `requirements.txt` file.

HenriqueAJNB:
Hi @jianboy. The requirements are defined at `pyproject.toml` and, as @gventuri said, the package uses `poetry` to manage virtual environments and dependencies.

PR #79, opened 2023-05-07
refactor: use contextlib.redirect_stdout to capture the output

walison17:
Replaces manual capture of output with the nice function [capture_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout)  from contextlib

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -1,6 +1,6 @@
 """ PandasAI is a wrapper around a LLM to make dataframes convesational """
 import io
-import sys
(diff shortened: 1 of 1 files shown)


gventuri:
Definitely better than the current implementation, thanks a lot @walison17! 🎉

avelino:
@gventuri if the implementation is better, why did you close the PR?

gventuri:
Guess I need a rest @avelino! Thanks for pointing it out, and @walison17 thanks a great for the big improvement!

PR #88, opened 2023-05-08
Refactor code for Cohere LLM (Need to improve it)

dSupertramp:
We actually don't have a response, so we should fix the response


Lorenzobattistela (commented):
Implementation looks correct and well done, I would suggest to create tests for your LLM (we're aiming to increase test coverage #89 ) and specify the version for cohere (to pass the CI, poetry is failing because of project consistency)

.gitignore
```diff
@@ -11,3 +11,11 @@ __pycache__
...
 dist
 pandasai.egg-info
+
+
+# virtualenv
+env
+ENV
+
+# test script
+test.py
```
gventuri, line 21: Maybe this is a leftover?

pandasai/llm/cohere.py
```diff
@@ -0,0 +1,111 @@
...
+    presence_penalty: float = 0.6
+    stop: str = None
+
+    def __init__(
+        self,
+        api_token: str = None,
+        **kwargs,
+    ):
+        self.api_token = api_token or os.getenv("COHERE_API_KEY")
+        if self.api_token is None:
```
gventuri, line 40: Could you also add this variable to the `.env.example` file?

gventuri (changes requested):
@dSupertramp thanks a lot for the contribution!
Did you test it? How does it work with the basic tasks provided in the example files?

.gitignore
dSupertramp: Ops

pandasai/llm/cohere.py
dSupertramp: Done!

dSupertramp:
> @dSupertramp thanks a lot for the contribution! Did you test it? How does it work with the basic tasks provided in the example files?

For now, we only have the basic response, without any kind of code or formatting: that's why we need to improve it

yzaparto:
@dSupertramp let me know if any help is required. Feel free to message me on discord

gventuri:
@dSupertramp, ok got it, thanks for clarifying!

What do you think is the reason why it doesn't return the correct code? Wrong prompt? Did you make some attempts? Let me know and let me know if we can help you with this!

PR #93, opened 2023-05-09
Automatically add plt.show() if missing, add error handling and fix bugs

### LINKED ISSUE #91: Starcoder Hallucinations, Graph Issues, Output issues
### 🐛 Describe the bug

When using `StarCoder`, it works great for text prompts 90% of the time, but graphing prompts don't seem to work. I have listed two issues I have observed and their respective prompts:

Prompt 1: Show a bar chart of the order qty for all unique part id's.
Screenshot 2023-05-08 at 12 55 58 PM
Issue 1: Besides not generating the `plt.show()` function, it seems to cut off randomly during the output. Don't know if this is a token issue, or some kind of execution issue. The `matplotlib` window appears in the MacOS Dock, so it is being initiated, but it doesn't open a window or anything.

Prompt 2: Make a bar chart of the order qty for all desired ship dates.
Screenshot 2023-05-08 at 1 00 19 PM
Issue 2: This prompt seems to make the LLM output quite an ambiguous message, and I don't see any code generated. 

I have attached the test data file 
[filename.csv](https://github.com/gventuri/pandas-ai/files/11423465/filename.csv)

Minimalized Code Is Below:
```python3
import pandas

from pandasai import PandasAI
from pandasai.llm.starcoder import Starcoder
        
dataPrompt = "Make a bar chart of the order qty for all desired ship dates."

llm = Starcoder(api_token="HF_API_TOKEN")

pandasAI = PandasAI(llm)

pandasDataFrame = pandas.read_csv("filename.csv")

response = pandasAI.run(pandasDataFrame, prompt = dataPrompt)

print(response)
```

yzaparto:
Changelog:

- Implemented the ability to add `plt.show()`, while plotting if it's not generated by llm.
- Added error handling if the LLM fails to generate code even after using the `error_correcting_mechanism`
- Fixes #91


pandasai/constants.py
```diff
@@ -73,3 +74,5 @@
     "vars",
     "zip",
 ]
+PLOT_KEY_WORDS = ["plot", "plt", "graph", "chart"]
+PLT_CODE = "\nplt.show()"
```
yzaparto, line 78: Should it be `"\nplt.show(blocking = False)"`

gventuri: @yzaparto totally agree!

pandasai/__init__.py
```diff
@@ -196,13 +199,17 @@ def run_code(
             # Execute the code
             count = 0
             code_to_run = self.remove_unsafe_imports(code)
+            _success_in_running_code = False
             while count < self._max_retries:
                 try:
+                    if any([x in code_to_run for x in PLOT_KEY_WORDS]) and code_to_run.count(PLT_CODE) < 1:
```
gventuri: What if it is already showing the plot? We show it again? remember it could show with `block= False` or not being passed, so let's make sure all the cases are considered

pandasai/__init__.py
```diff
@@ -240,11 +248,17 @@ def run_code(
                         error_correcting_instruction, ""
                     )
 
-        captured_output = output.getvalue()
+        if(_success_in_running_code):
```
gventuri: Let's try to avoid if/else when not necessary!

```
if not _success_in_running_code:
            raise ExecutionFailed("The llm could not generate the correct code.")

captured_output = output.getvalue()
```

gventuri:
Great job @yzaparto 🎉!
* Don't be afraid of all the comments left, only minor things!
* Just make sure to `pre-commit install` in order to avoid the indentation changes.
* Also, I'd ask you to write as many tests as possible to enforce what you've done, given the fact that what you just did is a very important and critical part!

gventuri:
@yzaparto did you manage to have a look at it?

Issue #104, opened 2023-05-10
Design of the CLI


victor-hugo-dc:
How would you like the design of the CLI to be? That is, would you want it to be a simple Python script that can execute a chunk of pandasai code given a prompt, dataset, api token, etc. or would you want something more complex? Would you want it to be a part of this repository or would you want to make the CLI a different repository altogether?

gventuri:
I'd say if it's not huge, we can leave it as part of this repo for the time being. Then if it becomes bigger and bigger we'd rather move it to a dedicate repo. @victor-hugo-dc what do you think?

victor-hugo-dc:
I added the command line tool as a part of the repo in my latest PR, it's fairly a small tool that simply requires `python setup.py install` to be run.

PR #107, opened 2023-05-10
CLI

victor-hugo-dc:
**Command-Line Interface Tool** 
From #104, I took the liberty to design a simple CLI tool that handles a variety of file type inputs (not just .csv) and can either take in a users token in the command line or can pull from the `.env` file directly. For the best user experience, I have set the default output of pandasai to be verbose and made sure to implement various error checks that print out only a single line detailing what went wrong.

Screenshot 2023-05-10 at 7 45 59 PM

### DIFF
--- a/pai/__main__.py
+++ b/pai/__main__.py
@@ -1,53 +1,46 @@
-import argparse
+import os
+from dotenv import load_dotenv
+import click
+import pandas as pd
 from pandasai import PandasAI
 from pandasai.llm.openai import OpenAI
 from pandasai.llm.open_assistant import OpenAssistant
-import pandas as pd
-
-import os
-from dotenv import load_dotenv
 
 dotenv_path = os.path.join(os.path.dirname(__file__), '..', '.env')
 load_dotenv(dotenv_path=dotenv_path)
 
-parser = argparse.ArgumentParser(description='Pandas AI command-line tool.')
-parser.add_argument('-d', '--dataset', type=str, required=True, help='The dataset to use.')
-parser.add_argument('-t', '--token', type=str, required=False, default=None, help='The API token to use.')
-parser.add_argument('-m', '--model', type=str, required=True, help='The type of model to use.')
-parser.add_argument('-p', '--prompt', type=str, required=True, help='The prompt to use.')
-
-def main():
-    args = parser.parse_args()
-
-    ext: str = os.path.splitext(args.dataset)[1]
+@click.command()
+@click.option('-d', '--dataset', type=str, required=True, help='The dataset to use.')
+@click.option('-t', '--token', type=str, required=False, default=None, help='The API token to use.')
+@click.option('-m', '--model', type=click.Choice(['openai', 'open-assistant']), required=True, help='The type of model to use.')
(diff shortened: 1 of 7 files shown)


pai/__main__.py
```diff
@@ -0,0 +1,57 @@
...
+from pandasai.llm.open_assistant import OpenAssistant
+import pandas as pd
+
+import os
+from dotenv import load_dotenv
+
+dotenv_path = os.path.join(os.path.dirname(__file__), '..', '.env')
+load_dotenv(dotenv_path=dotenv_path)
+
+parser = argparse.ArgumentParser(description='Pandas AI command-line tool.')
```
avelino, line 13: in the case of cli we could use the [click package](https://click.palletsprojects.com/en/8.1.x/) to simplify the implementation and maintenance - the code is more organized

avelino (changes requested):
given that we are making a cli we need to think about package distribution for macOS, Linux and Windows, given the option for people to install in a "simple" way - not only via pip, e.g. homebrew for macOS

pai/__main__.py
```diff
@@ -0,0 +1,50 @@
...
+    except:
+        print(f"Could not find {dataset}")
+        return
+
+    if model == "openai":
+        llm = OpenAI(api_token = token or os.environ.get("OPENAI_API_KEY"))
+
+    elif model == "open-assistant":
+        llm = OpenAssistant(api_token = token or os.environ.get("HUGGINGFACE_API_KEY"))
+
```
gventuri, line 40: Let's also add `Starcoder`, which we are now supporting!

pyproject.toml
```diff
@@ -15,6 +15,7 @@ astor = "^0.8.1"
 openai = "^0.27.5"
 ipython = "^8.13.1"
 matplotlib = "^3.7.1"
+click = "^8.1.3"
```
gventuri, line 18: Isn't it a dev dependency?

gventuri (changes requested):
Great job @victor-hugo-dc! 🎉
I'll also give it a try and let you know how it feels!

pyproject.toml
victor-hugo-dc: I'm still fairly new to contributing to open source projects, I'm unsure if its a dev dependency; should i move it from dependencies to dev dependencies?

HenriqueAJNB:
I agree with you in parts @avelino. What you suggested is important, but first the package must have the CLI implemented.

Your proposal is a nice to have in future versions. Perhaps you could even open another issue to track it.

avelino:
> I agree with @avelino. What he suggested is important, but first the package must have the CLI implemented.
> 
> His proposal is a nice to have in future versions. Perhaps @avelino could open another issue to track it.

@HenriqueAJNB
"could use": **recommendation** to use, not ~~forcing~~

gventuri:
@victor-hugo-dc tested, works super cool.

2 things left:
- fix the `linter` errors
- add an example of how to set it up and how to run it in the `README.md`

Great job @victor-hugo-dc 🎊

PR #112, opened 2023-05-10
feat: add auto pypi publish on new release

### LINKED ISSUE #101: Automate pypi release in github actions
### 🚀 The feature

I noticed that currently the release process to pypi is manually triggered. This however can be automated whenever a new release is published. 

The trigger of such workflow can be found e.g. [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release).

Regarding the release job, it can either be written by utilising the [pypi-publish](https://github.com/marketplace/actions/pypi-publish) github action, or you can formulate the build steps following the [guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) from PyPA.

Please let me know if this is something we are interested, and I would also love to contribute a PR if needed. (though some helps may needed from admins regarding credentials and test runs).

### Motivation, pitch

I noticed that currently the release process to pypi is manually triggered. This however can be automated whenever a new release is published.

### Alternatives

_No response_

### Additional context

_No response_

TiansuYu:
- [X] closes #101
- [X] Tests added and passed if fixing a bug or adding a new feature

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,28 @@
+name: pandas_ai_cd
+
+on:
+  releases:
+    types:
+      - published
+  # for testing without release a new version
+#  push:
+#    branches:
+#      - add-auto-pypi-release
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  publish_to_pypi:
+    name: publish to pypi on new release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: JRubics/poetry-publish@v1.16
+        name: Build and publish to Pypi
+        with:
+          pypi_token: ${{ secrets.PYPI_TOKEN }}
+          ignore_dev_requirements: "yes"
+          repository_url: https://test.pypi.org/legacy/ # for testing only
+          repository_name: test_pypi # for testing only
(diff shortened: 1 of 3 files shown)


TiansuYu:
@gventuri I have written a draft. If you want to test it, you can go add a test pypi person access token as `PYPI_TOKEN` in github secret (which can be replaced with real ones afterwards) and switch on this section:
```
#  push:
#    branches:
#      - add-auto-pypi-release
```
to test the auto-publish. 
(Unfortnately, I think the trigger on release will not work unless this yaml is on main.) 

After this works, I will remove the testing part and leave only the parts for production.

gventuri:
@TiansuYu configured everything, but cannot make it work:

`There is no pypi-token.test_pypi setting.`

This is actually weird, cause we are passing the key

gventuri:
@TiansuYu if what the maintainer of the library mentioned is correct, maybe we should test with an actual release with the right repo and see whether it works!

HenriqueAJNB:
@gventuri, at the end of this [video](https://youtu.be/0f3moPe_bhk) there is how to publish a package with poetry that could be allocated in CI. 

By the way, I'm the one who made the outline and examples for that vídeo.

TiansuYu:
I will come back to this as soon as I can. In the mean time, the PR is not correct from my point of view. 

Regarding the action itself. 1) fix according to the action author's suggestions 2) we can write the steps ourselves without the action's help. It is something super simple with github.

Issue #96, opened 2023-05-09
Support for Azure Openai apis


reigN0ne:
### 🚀 The feature

Hi, was wondering if we can get the support for azure openai directly as we have for other llms? Is something happening under development for the cause?

### Motivation, pitch

Azure Openai is a great way to get used to the LLM technology without having to deal with anything else. Hence would love to see the features from pandas-ai implemented for the same, which will help greatly towards creating analysis reports.

yzaparto:
Sounds like a great idea. @reigN0ne is the Azure open ai available to regular users yet? It looks like they are giving access to enterprise customers.

<img width="780" alt="image" src="https://user-images.githubusercontent.com/132240560/237028345-afd1d729-79bc-4cb9-9536-a5a77913d74c.png">

UnderTheCraft:
Would be interested as well.
@yzaparto AFAIK, you might apply to it via [this form](https://customervoice.microsoft.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbR7en2Ais5pxKtso_Pz4b1_xUOFA5Qk1UWDRBMjg0WFhPMkIzTzhKQ1dWNyQlQCN0PWcu). But I'm not sure, how long you'll need to wait, since I got access via my company account.

Regarding the implementation, we could check LangChain's version of AzureOpenAI for this feature (see https://github.com/gventuri/pandas-ai/blob/main/pandasai/llm/openai.py).

mspronesti:
I guess this issue can be closed now :)

gventuri:
@mspronesti nice catch, closing!

PR #115, opened 2023-05-11
Support for Azure Openai

mspronesti:
Hello,
this PR aims at integrating Azure OpenAI API into pandasAI, addressing #96 .

I went for separate implementation instead of adding the required changes to `openai.py` for the following two reasons:
- keeping the latter cleaner, avoiding ambiguities between the model, the engine and the additional params Azure requires, which would be all ignored in the "classical" OpenAI version
- it's not so straightforward in Azure OpenAI to extract the model name from the deployment name and to associate it to a completion or a chat model


gventuri (changes requested):
Great job! I see much code duplication though. What do you think about having a common class that is then extended by both OpenAI and AzureOpenAI?
Or even better: AzureOpenAi could extend OpenAI and only override the methods that need to be overridden.

gventuri:
Great job @mspronesti 🎉 Just left you a minor suggestion in terms of architecture!

gventuri:
Great job @mspronesti 🎉 Just left you a minor suggestion in terms of architecture!

mspronesti:
I'd suggest going for the first option rather than for the second, as every method would need to be overwritten in that case. In fact, there are two key points that forced me to change "at least something" in every method:
- Azure OpenAI requires an `engine` or a `deployment_id`, not a `model` 
- I couldn't find a way to discriminate between chat and completion models given a deployed model in Microsoft Azure, as one can even multiple deployments of the same model with different names. Thus the need to have a different `call`. 

I can try to design a `BaseOpenAI` class to avoid some duplicates here and there.

gventuri:
@mspronesti yeah, then I'd say go for it, I like the approach!

mspronesti:
Sure, thanks for the feedback :) just one last question: is `call` not decorated with `abstractmethod` on purpose ?

gventuri:
@mspronesti no, it actually makes sense to decorate it with `abstractmethod`! Would be great if you could do it!

mspronesti:
@gventuri I just appended a commit to include a base class and generalize the behavior. In my next commit I will add the decorators, thanks for your reply :)

mspronesti:
@gventuri  done :)
I also discovered that on Microsoft Azure you can map the deployment name into the model name quite easily, but it takes 2 lookups:
- search the list of objects to extract the model name
- check whether the model name is in some `_supported_completions_models` or `_supported_chat_models` lists.

```py
deployments = openai.Deployment.list()
model_name = [deployment for deployment in deployments.data if deployment.id == deployment_name][0].model

if model_name in _supported_completions_models:
    ...
elif ...
```

in my view, given the fact that the user knows what his model is for, it's way more efficient to let him set the `is_completion_model` parameter (as in the current version). However, if you prefer the option I mentioned above, let me know and I will change this too :)

UnderTheCraft:
nice 👍

pandasai/llm/base.py
```diff
@@ -107,3 +113,89 @@ def generate_code(self, instruction: str, prompt: str) -> str:
             str: Code
         """
         return self._extract_code(self.call(instruction, prompt, suffix="\n\nCode:\n"))
+
+
+class BaseOpenAI(LLM, ABC):
```
gventuri, line 118: I'd say let's either create a `base_openai.py` or move also the `base_hf.py` here, so that all the "base" models are grouped. What do you think?

gventuri:
> @gventuri Done. It should be ready to go :) I also discovered that on Microsoft Azure you can map the deployment name into the model name quite easily, but it takes 2 lookups:
> 
> * search the list of objects to extract the model name
> * check whether the model name is in some `_supported_completions_models` or `_supported_chat_models` lists.
> 
> ```python
> deployments = openai.Deployment.list()
> model_name = [deployment for deployment in deployments.data if deployment.id == deployment_name][0].model
> 
> if model_name in _supported_completions_models:
>     ...
> elif ...
> ```
> 
> in my view, given the fact that the user knows what his deployed model is for, it's way more efficient to let him set the `is_completion_model` parameter (as in the current version). Besides, the supported models would be sandbox-specific and not available in every continent. However, if you prefer the option I mentioned above, let me know and I will change this too :)

Great job @mspronesti! Let's leave it as it is for now, I like it better. If it becomes hard to maintain we can change it at a later point (we are doing the same with OpenAI as of now, only supporting some models).

pandasai/llm/base.py
mspronesti: I'd prefer having everything in `base.py`, but it's quite subjective. Since they all are base classes for the "backend" of pandasAI and they are small classes, I believe a unique file would serve the purpose. However the other way around would be fine as well in my opinion so you tell me :)

gventuri:
One final thing, then we are good to go. @mspronesti could you write some tests (check out the other LLMs). We are trying to increase the coverage, as the library is gaining more and more traction.

Thanks a lot for the great work @mspronesti, super excited to have it merged!

mspronesti:
Sure @gventuri, I will add some unit tests in the next commit. 
Let me know also what you prefer for base classes file management so that I can fix that too :)
Thanks a lot for the review!

pandasai/llm/base.py
mspronesti: Do you want me to move the content of `test_base_hf.py` inside `test_base.py` as well ?

gventuri:
> Sure @gventuri, I will add some unit tests in the next commit. Let me know also what you prefer for base classes file management so that I can fix that too :) Thanks a lot for the review!
> 
> EDIT: except for the parameters testing, all the completion and chat tests could be copied and pasted since the logic of the two APIs is the same. Any suggestion on this ?

I'd say for now let's ship it with some code duplication for the sake of simplicity. But we definitely need to think about how to avoid code duplication here.

mspronesti:
would you prefer them in a separate file or in `test_openai.py` ? Also, given the fact that the real functionalities are shared in the base class, I believe it only makes sense to re-test the constructor in the various scenarios (parameters, exceptions). For this reason I'm considering, if you agree, to put them in the very same test file. What do you think ? :)

mspronesti:
Hello @gventuri, 
I have another update for this PR. Upon further study of the API, there's the chance to also do this

```py
model_id = openai.Deployment.retrieve(deployment_name).model # e.g. gpt-35-turbo
capabilities = openai.Model.retrieve(model_id).capabilities
"""
<OpenAIObject at 0x2364df244a0> JSON: {                                                                                         
 "chat_completion": true,                                                                                                       
 "completion": true,                                                                                                            
 "embeddings": false,                                                                                                           
 "fine_tune": false,                                                                                                            
 "inference": true,                                                                                                             
 "scale_types": [                                                                                                                
    "standard"                                                                                                                   
    ]                                                                                                                            
  }   
"""
```

the last part could be even ported in the current openai model, so that we don't have to hardcode the supported models and distinguish "manually" between chat and completion, thus letting the user use whatever model he wants
In principle, with this approach one might even considering having a unique class both for openai and azure openai. What do you think ?

pandasai/llm/base.py
gventuri: Would be great! @mspronesti

gventuri:
@mspronesti totally! Go for it :)

mspronesti:
@gventuri sure :) in the meanwhile, upon further study of the API, I changed a bit the logic, auto-deducing the model capability. Let me know if you like this solution more than the previous version.

mspronesti:
@gventuri Tests added. I eventually opted to put them in a separate file as I had to mock quite some API call and test some more scenarios that would have made a single file difficult to maintain, in my opinion. This PR should be ready to go now :)

gventuri:
@mspronesti great job! We are almost there, but the linting is faling (https://github.com/gventuri/pandas-ai/actions/runs/4973531138/jobs/8900957487?pr=115). Could you fix it, the I'll merge it?

mspronesti:
Done! Sorry for not noticing the lint failure :/ I only left two warnings complaining about non-explicitly re-rasing the exception. Not sure if this would make the workflow fail :)

mspronesti:
Done @gventuri . Now all the checks should pass :)

gventuri:
@mspronesti great, merging it! 🚀

ellenealds:
Hey guys, I am trying to use the AzureOpenAI function but I am getting an error as described below.

Platform: Databuton
Error: ModuleNotFoundError: No module named 'pandasai.llm.azure_openai'

Installed packages:: 
databutton==0.31.9
streamlit==1.22.0
ipykernel==6.22.0
watchdog==3.0.0
openai
pandasai
astor
python-dotenv

Page installs:
import databutton as db
import streamlit as st
import pandas as pd
import pandasai
import openai
from pandasai.llm.azure_openai import AzureOpenAI
from pandasai import PandasAI

Error:

ModuleNotFoundError: No module named 'pandasai.llm.azure_openai'
Traceback:
File "/user-venvs-build/ec7e6b749188/.venv/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 565, in _run_script
    exec(code, module.__dict__)
File "/app/run/multipage/Home.py", line 6, in <module>
    from pandasai.llm.azure_openai import AzureOpenAI

mspronesti:
@ellenealds which version of pandasai have you installed ? 
```shell
pip list | grep pandasai
```

ellenealds:
This is really strange, I get this error when I try to check it:

AttributeError: module 'pandasai' has no attribute '__version__'
Traceback:
File "/user-venvs-build/ec7e6b749188/.venv/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 565, in _run_script
    exec(code, module.__dict__)
File "/app/run/multipage/Home.py", line 5, in <module>
    st.write("PandasAI version:", pandasai.__version__)

gventuri:
@ellenealds which version of python do you have installed? Remember, PandasAI only supports python 3.9+

mspronesti:
@ellenealds you can check `pandasai`'s version with the command I wrote in my previous comment.
On Linux/MacOS

```shell
pip list | grep pandasai
```

on Windows

```shell
pip list | Select-String pandasai 
```

@gventuri from the log, I guess it's python3.10

ellenealds:
I am am using Databutton which is browser based, I have tried my code in VS and this works perfectly. I will speak to the Databutton team as I am writing up a blog post for PandasAI using Databutton. I may have some questions from them but I am sure that there is some sort of compatibility issue on their end. VERY excited for this, really amazing library thank you :)

Issue #53, opened 2023-05-05
Google Sheets Feature request


stikkireddy:
Hey can you use gspread or other frameworks so that we can run some cool prompt based analytics for pandas.

gventuri:
Sounds like a very exciting feature @stikkireddy, will look into it!

aryandgandhi:
Hey @gventuri I would love to look into this

gventuri:
Hey @aryandgandhi, sounds super! Just assigned to you!

aryandgandhi:
Awesome, just wanted to clarify @gventuri and @stikkireddy is the goal to pull in the google sheets data into a dataframe and run specific prompts, Ex. create a graph where each state is a different color?

gventuri:
Yeah, would be a great starting point. Like we pull data from google sheets and convert to a dataframe that then we pass to PandasAI! @aryandgandhi

gventuri:
Closing, we now have a different and more comprehensive issue for this #301

PR #120, opened 2023-05-13
added google sheets functionality

aryandgandhi:
Hello,
this PR aims at integrating Google Sheets into pandasAI, addressing https://github.com/gventuri/pandas-ai/issues/53 .

I wrote a file where the user sees an example of getting their sheetid and name as inputs, and can enter into into the sheets_input() function to use their sheets dataframe

*used an environment variable for sheet_id as the spreadsheet is public and did not want any changes to be made, however it uses the same gdp data that is already used in the project


gventuri:
Hello @aryandgandhi, great job! Only a couple of minor improvements:
* can you create a helper `from_google_sheets` that returns a df that we can then pass to pandas ai? This would further facilitate the job. What I mean is, instead of declaring in the example, saving it as a helper and easily importing in the example.
* don't forget to write the tests for the helper
* could you also add a env variable in `env.example`?

Thanks a lot for your support!

aryandgandhi:
On it @gventuri!

gventuri:
@aryandgandhi thanks a lot! However, there are a couple of things we need to address:
* add a way to somehow login with the google sheets api token (otherwise we cannot be sure the script will work in every environment)
* we don't need a variable GOOGLE_SHEETS in the `.env`. Just need to store the `GOOGLE_SHEETS_API_KEY` that we'll use to authenticate

We are almost there, let's just make it bullet proof

aryandgandhi:
Hey @gventuri let me know what you think

pyproject.toml
```diff
@@ -15,6 +15,13 @@ astor = "^0.8.1"
 openai = "^0.27.5"
 ipython = "^8.13.1"
 matplotlib = "^3.7.1"
+google-auth = "^2.18.0"
```
avelino, line 18: we are forcing to install Google library even if one does not use Google Sheets, this is a problem for the future of the **pandasai** project, we will have N LLM and multiple output and input support (from various projects), this is not healthy for **pandasai**

we can use [extra pip dependency](https://peps.python.org/pep-0508/#extras), [poetry documentation](https://python-poetry.org/docs/pyproject/#extras)

cc @gventuri

avelino (changes requested):

examples/from_googlesheets.py
```diff
@@ -0,0 +1,26 @@
+import pandas as pd
+from data.sample_dataframe import dataframe
+
+from pandasai import PandasAI
+from pandasai.helpers.datasource import sheets_input
+from pandasai.llm.openai import OpenAI
+
+# instructions: https://docs.google.com/document/d/1uAmXI3e8Wm26lZQb0bYIJaXRdZi5WLJKNJHT59QMGEc/edit?usp=sharing
+df = sheets_input(
```
gventuri, line 9: Let's call it `from_google_sheets` in a similar way of what we are doing for csv (`from_csv` and xls `from_excel`). What do you think?

examples/from_googlesheets.py
```diff
@@ -0,0 +1,26 @@
...
+    "1IRGJFUpCR0-9tPDk2MgPjQCtNft_4a8B1fHgYK7_LM8",
+    "second",
+    "../service_account_file.json",
+)
+# specify type of columns, google sheets is not strongly typed
+df["gdp"] = pd.to_numeric(df["gdp"])
+df["happiness_index"] = pd.to_numeric(df["happiness_index"])
```
gventuri, line 16: This should be automatically done by default as we run `from_google_sheets `, what do you think? Otherwise having to manually make this transformation all the times could be very repetitive.

pyproject.toml
gventuri: @avelino, I agree, we should find a compromise to keep it lightweight!

@aryandgandhi what do you think about adding it in `[tool.poetry.extras]` instead? Check out [poetry documentation](https://python-poetry.org/docs/pyproject/#extras)

tests/test_datasource.py
```diff
@@ -0,0 +1,24 @@
...
+        assert (
+            sheets_input("1IRGJFUpCR0-9tPDk2MgPjQCtNft_4a8B1fHgYK7_LM8", "second")
+            is not None
+        )
+
+    #  def correct_output(self):
+    #     assert (
+    #         sheets_input("1IRGJFUpCR0-9tPDk2MgPjQCtNft_4a8B1fHgYK7_LM8", "second")
+    #         is not None
+    #     )
```
gventuri, line 24: Let's remove this

service_account_file.json
```diff
@@ -0,0 +1,3 @@
+{
+}
```
gventuri, line 2: Should this be committed?

pandasai/helpers/datasource.py
```diff
@@ -0,0 +1,43 @@
...
+    # pylint: disable=no-member
+    sheet = service.spreadsheets()
+    result = (
+        sheet.values()
+        .get(spreadsheetId=spreadsheet_id, range=spreadsheet_name)
+        .execute()
+    )
+
+    values = result.get("values", [])
+    print(values)
```
gventuri, line 40: Let's remove this print or make it printable only if a specific param is passed!

gventuri:
@aryandgandhi impressive job, just left you a couple of comments. Looking forward to merge this one!

aryandgandhi:
Hey @gventuri, check #180 there was an issue with committing, the google sheets should be ready now, just wanted to make sure the package listings in tool.poetry.extras made sense

aryandgandhi:
Hey @gventuri should be all done, go ahead and check the project dependences to see if that works

.gitignore
```diff
@@ -15,11 +15,18 @@ pandasai.egg-info
 
 #venv
 /venv
+<<<<<<< HEAD
+=======
 .venv
+>>>>>>> upstream/main
```
gventuri, line 21: Here there's a issue with handling the conflict

examples/from_googlesheets.py
```diff
@@ -0,0 +1,27 @@
+"""Example of using PandasAI with GoogleSheets"""
+
+import pandas as pd
+from data.sample_dataframe import dataframe
+
+from pandasai import PandasAI
+from pandasai.helpers.datasource import from_google_sheets
+from pandasai.llm.openai import OpenAI
+
+# instructions: https://docs.google.com/document/d/1uAmXI3e8Wm26lZQb0bYIJaXRdZi5WLJKNJHT59QMGEc/edit?usp=sharing
```
gventuri, line 10: Great job with the instructions. Would you mind adding them as part of the documentation? You can find it in the `/docs` folder!

examples/from_googlesheets.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+df = from_google_sheets(
+    "1IRGJFUpCR0-9tPDk2MgPjQCtNft_4a8B1fHgYK7_LM8",
+    "second",
+    "../service_account_file.json",
```
gventuri, line 16: I would probably set this as an optional step with a default value (which could be `google_service_account_file.json` by default. What do you think?

service_account_file.json
gventuri: I suggest we rename it to `google_service_account_file.json`. What do you think?

gventuri (commented):
Impressive job, almost there!

gventuri:
@aryandgandhi also make sure you update the dependencies. There is some kind of conflict: https://github.com/gventuri/pandas-ai/actions/runs/5100391674/jobs/9173884219?pr=120

Finally make sure both the linter and the tests pass!

PR #123, opened 2023-05-14
Fix bug using markdown format. print(df.head({rows_to_display})) 

zhenkaivip:
Fix bug using markdown format. If max_cols is exceeded, switch to truncate view.


gventuri:
Hey @zhenkaivip, thanks a lot for the MR.
I noticed you haven't included the `tabulate` dependency.
I'm not even sure we need a new dependency for that. What if we do it manually with a helper (like it column > 10, truncate it)? We could, in the future, find a way to also let the user customize the max columns field.

zhenkaivip:
we can set max_columns like this :
```
pd.set_option('display.max_columns', None)
```
But it will wrap, split a row into multiple rows, and connect them with ```\```. This is not a beautiful format.
So the markdown format is a better solution.
I'm sorry I forgot to tell you  ```pip install tabulate```. I have already installed it before.
You can test it using an Excel with many columns.

gventuri:
@zhenkaivip in order to keep it as clean as possible (and not having to add additional dependencies), I would recommend the following approach:

* right before we pass the prompt, we set `pd.set_option("display.max_columns", 10)`
* right after the prompt, we set `pd.set_option("display.max_columns", None)`

P.s. please, remove gradio, which is not part of this PR!

Great idea btw @zhenkaivip, thanks a lot for this PR!

PR #154, opened 2023-05-21
feat: add from_excel (#121)

### LINKED ISSUE #121: Add Excel Functionality
### 🚀 The feature

Be able to read in an excel dataframe

### Motivation, pitch

I'm not a great programmer, want the ability to read in data from excel

### Alternatives

N/A

### Additional context

N/A

jonbiemond:
Add from_excel function, a limited and easy-to-call pd.read_excel() wrapper.

- [x] closes #121
- [x] All [code checks passed]


jonbiemond:
@gventuri I have a couple notes for your review.

- File organization: please review how I organized the new files.
- Example data: I was hesitant to add a new .xlsx file to the project, so in the example I reuse the .csv, feels janky.
- Permission error handling: A common error with reading Excel files is a permission error when the user has the file open in Excel. This might be hard for an unfamiliar user to understand. I could catch, and raise something more verbose.

pandasai/helpers/from_excel.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Return a pandas DataFrame from an Excel file.
+    Wrapper for pandas.read_excel().
+
+    :param file: str, Path to the Excel file to read from.
+    :param sheet: str, int, None, default = 0 Name of the sheet to read in,
+        or the sheet number (0-indexed). Defaults to the first sheet.
+    :return: pd.DataFrame, Dataframe containing the data from the specified sheet.
+    """
+    return pd.read_excel(file, sheet_name=sheet)
```
mspronesti, line 17: What is the point of a util that only wraps a single call to pandas function without any additional logic ? one can simply use `pd.read_excel(.)` before passing the df to pandasai, can't he ?

jonbiemond: Abstraction. See the issue #121
But I agree it's can be seen as superfluous and depends on the overall aim of the project.

examples/from_excel.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pandas as pd
+
+from pandasai import PandasAI
+from pandasai.helpers.from_excel import from_excel
+from pandasai.llm.openai import OpenAI
+
+# convert csv to xlsx for this example
+df = pd.read_csv("examples/data/Loan payments data.csv")
+df.to_excel("examples/data/Loan payments data.xlsx", index=False)
```
gventuri, line 11: I would say let's rather add a `.xlsx` file in the `/data` folder and include that one. All these transformation make the example less intuiteve!

pandasai/helpers/from_excel.py
gventuri: Although I totally agree with @mspronesti that might not make 100% sense as of now, it's also true that having some abstraction make it even easier to use PandasAI for non-pandas users.
Kind of make sense to me, would be interesting to also have something like that for csv, for example, so that we can use it as easy as:

```
from pandasai import PandasAI
from pandasai.helpers.from_excel import from_excel

df = from_excel("examples/data/Loan payments data.xlsx")

llm = OpenAI()
pandas_ai = PandasAI(llm, verbose=True)
response = pandas_ai(df, "How many loans are from men and have been paid off?")
print(response)
```

Not a game changer, but it's nice to have, and might be useful for future developments. I don't know, don't have a strong opinion, what do you think @mspronesti @jonbiemond?

jonbiemond: Another option is making a `from_file` helper which can handle a variety of files and read them in according to the extension. For an example see the implementation from @victor-hugo-dc here: https://github.com/victor-hugo-dc/pandasai-ui/blob/917881a4b2e3da50002a2260e3f56f4fc5bb9b70/src/main.py#LL10C7-L10C7

mspronesti: My two cents: a pandas-ai user is to some extent also a pandas user because he has to, at least, be able to define a dataframe. Adding this function and documenting it is no different than simply adding a paragraph in the doc where you show how to read a `.xlsx` using plain pandas.

On the other hand, the `from_google_sheets` util at least has some logic in it, thus I find it useful. Here we are only renaming `pd.read_excel`  to `pandasai.from_excel`. 

Up to you though :)

jonbiemond: > Another option is making a from_file helper which can handle a variety of files and read them in according to the extension.

@mspronesti Thoughts on this?

mspronesti: @jonbiemond I'm subjectively skeptical about this option too. My point is that pandas supports many formata, among which excel, CSV, JSON, pickle, etc. All of this can be loaded with 1 line of pandas code. What would be the value of a util that only does an if-then-else ? I might be wrong, but I believe this lib's entrypoint should be a dataframe and the user should be at least able to create one. 
Again, up to know guys though, these are just my 2 cents :)

gventuri: @jonbiemond @mspronesti all in all, I guess the abstraction makes sense, so that we have many methods, like `from_csv`, `from_excel`, `from_google_sheets`. Although I agree this is not adding anything on the table right away, I guess it could be beneficial in the long run with further developments that will come soon.

pandasai/helpers/from_excel.py
```diff
@@ -0,0 +1,17 @@
+"""Functions to return a dataframe from an Excel file."""
```
jonbiemond, line 1: Note: PR #120 Uses a `datasource.py` module for the Google Sheets function. Whereas here a more specific module is used.

PR #168, opened 2023-05-25
feat: multiple dataframes

victor-hugo-dc:
I still need to fix up some things for pylint and add some tests for checking functionality. But this is the code for multiple dataframes support so far.


gventuri (commented):
This is by far the most exciting PR I've seen since injection. @victor-hugo-dc great job, it's amazing!

A couple of points:
- let's mention this feature also in the README file, so exciting to be kept in the dark
- let's write the tests

Looking forward to shipping this in a new release!

Great job

PR #171, opened 2023-05-25
chore(deps): make google-palm optional dep

### LINKED ISSUE #153: Limit the number of dependencies
### 🚀 The feature

The core idea would be to make the user install third party additional dependencies according to the engine LLM they want to use - except OpenAI, installed by default.

### Motivation, pitch

Recently Google Palm has been integrated into PandasAI. As a consequence, `google-generativeai` has been added to this projects deps. Now I'm working on integrating GPT4All and LlamaCpp (this one does not really produce good code though), which require, respectively, `gpt4all` and `llama-cpp-python`. 
In the end, a user interested in using only one engine, say OpenAI, would be forced to install all these unneeded additional deps.

### Proposed solution
The proposed solution would be to let the user install a dependency manually only when needed. Example:
```py
try:
      from gpt4all import GPT4All as GPT4AllModel
      ...
except ImportError:
        raise ...
```
An additional possibility would be to add all of these as extra deps to make pandasai installable as follows
```shell
$ pip install pandasai["llamacpp", "googleai", ...]
```
but I believe this is not really necessary.

If you feel this is a good idea, feel free to assign me to this issue :)

mspronesti:
Hi @gventuri, 
this PR aims at addressing #53. 

- [X] closes #153
- [X] All [code checks passed](https://github.com/gventuri/pandas-ai#contributing).

## Changelog:
- google-palm becomes an optional dep, installable via `poetry install --extras google` or `pip install pandasai[google]` or `poetry install --all-extras`. Reference: [poetry documentation](https://python-poetry.org/docs/pyproject/#extras)
- the test workflow installs also the optional dependencies to appropriately run all the tests. A possible alternative would be to instead use `pytest.importskip`, which is the approach adopted in langchain. This would skip the tests of the optional deps. I let you decide what you prefer :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install poetry
         run: pip install poetry==1.4.2
       - name: Install dependencies
-        run: poetry install
+        run: poetry install --all-extras
       - name: Lint with pylint
         run: poetry run pylint pandasai examples
       - name: Test with pytest
--- a/pandasai/llm/base.py
+++ b/pandasai/llm/base.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Optional
 
 import openai
 import requests
-from google import generativeai
 
 from ..constants import END_CODE_TAG, START_CODE_TAG
 from ..exceptions import (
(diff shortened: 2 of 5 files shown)


gventuri:
@mspronesti great job, also love the approach of installing everything to write the tests. We want to ensure all the codebase (including the optional parts) are well tested!

Merging!

PR #179, opened 2023-05-27
fix(hf): wrong concatenation of prompt and suffix

### LINKED ISSUE #178: Huggingface model's no longer working after introducing Prompt classes
### 🐛 Describe the bug

Hi @gventuri,
this is a minor bug affecting Hugginface LLMs. When using one as a backbone for `PandasAI` I get

```

unsupported operand type(s) for +: 'GeneratePythonCodePrompt' and 'str'
```

This is due to a minor mistake in the way the prompt and the suffix are concatenated after the `Prompt` class (and its derivations) have been introduced.

mspronesti:
- [X] closes #178  
- [X] All [code checks passed](https://github.com/gventuri/pandas-ai#contributing).

### DIFF
--- a/pandasai/llm/base.py
+++ b/pandasai/llm/base.py
@@ -226,7 +226,8 @@ class HuggingFaceLLM(LLM):
     def call(self, instruction: Prompt, value: str, suffix: str = "") -> str:
         """Call the LLM"""
 
-        payload = instruction + value + suffix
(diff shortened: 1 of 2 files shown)


gventuri:
@mspronesti great job, I've just added a test to enforce the right behavior (I wish I did before).

PR #181, opened 2023-05-27
feat: multiple dataframes

victor-hugo-dc:
Closing the other one because I made a mistake in merging and I'm bad at git so I couldn't figure out how to fix it. Can you help me write some tests? I'm not exactly sure what we should be testing for.


gventuri:
@victor-hugo-dc amazing job. Added the tests and added it on the docs, merging!

PR #193, opened 2023-05-31
feat: whitelist of optional dependencies

### LINKED ISSUE #191: Generated code with library not in whitelist fails silently
### 🐛 Describe the bug

When the code generated includes references to a library not in the whitelisted libraries it simply fails with no indication as to why. I would expect something like an ImportError to be raised.

### Example
Generated code:
```python
import pandas as pd
import seaborn as sns

# Load the example dataset for Anscombe's quartet
df = sns.load_dataset("anscombe")

# Show the results of a linear regression within each dataset
sns.lmplot(
    data=df, x="x", y="y", col="dataset", hue="dataset",
    col_wrap=2, palette="muted", ci=None,
    height=4, scatter_kws={"s": 50, "alpha": 1}
)
```
Results in: `""`

### 🚀 Request
At this point I'll turn my bug report to a feature request. I think it would be great if we have a feature that allows the user to install the missing library, add it to the whitelist and re-run the code.

jonbiemond:
Fixes bug by raising an exception when generated code contains an import for a library that isn't whitelisted.
Adds a feature to include a whitelist of optional dependencies. When generated code contains an import for a library on the optional whitelist but is not installed, raise an ImportError with an instructional message. 

- [x] 🐛 Raise exception when generated code contains an import for a library that isn't whitelisted.
- [x] 🚀 Add whitelist of optional dependencies.
- [x] closes #191


gventuri:
@jonbiemond impressive job!! Well done

PR #211, opened 2023-06-02
add libs to whitelist

### LINKED ISSUE #210: code not generated due to missing libs in whitelist
### 🐛 Describe the bug

After the latest release, the example did not generate code due to missing libs in whitelist,
```
import pandas as pd
  ...: 
  ...: from pandasai import PandasAI
  ...: from pandasai.llm.openai import OpenAI
  ...: 
  ...: from examples.data.sample_dataframe import dataframe
  ...: 
  ...: df = pd.DataFrame(dataframe)
  ...: 
  ...: llm = OpenAI()
  ...: pandas_ai = PandasAI(llm, verbose=True, conversational=False)
  ...: response = pandas_ai(df, "Calculate the sum of the gdp of north american countries")
  ...: print(response)
  ...: 
Running PandasAI with openai LLM...
Code generated:
import pandas as pd
# create dataframe
data = {'country': ['United Kingdom', 'France', 'Italy', 'Germany', 'United States', 'Canada', 'Mexico', 'Brazil', 'Argentina', 'Chile'],
        'gdp': [1745433788416, 7191747985, 5659112121, 2655246018, 4538789365, 1785387456, 1218745698, 2145789654, 987456321, 789456123],
        'happiness_index': [7.16, 7.07, 6.38, 6.38, 7.16, 7.23, 6.77, 6.95, 6.08, 6.44]}
df = pd.DataFrame(data)
# calculate sum of gdp for North American countries
north_america = ['United States', 'Canada', 'Mexico']
sum_gdp = df[df['country'].isin(north_america)]['gdp'].sum()
print(sum_gdp)
                
Generated code includes import of pandas which is not in whitelist.
```

sandiemann:
- [ ] closes #210 

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -323,7 +323,10 @@ class PandasAI:
                 if alias.name in WHITELISTED_OPTIONAL_LIBRARIES:
                     import_optional_dependency(alias.name)
                     continue
-                if alias.name not in WHITELISTED_LIBRARIES:
+                if not any(
+                    alias.name.startswith(whitelisted_lib)
+                    for whitelisted_lib in WHITELISTED_LIBRARIES
+                ):
                     raise BadImportError(alias.name)
 
         return False
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -318,20 +318,12 @@ class PandasAI:
 
         if isinstance(node, (ast.Import, ast.ImportFrom)):
             for alias in node.names:
-                if alias.name in WHITELISTED_BUILTINS:
+                if alias.name in WHITELISTED_BUILTINS or alias.name == "pandas":
                     return True
                 if alias.name in WHITELISTED_OPTIONAL_LIBRARIES:
                     import_optional_dependency(alias.name)
                     continue
-                if (
-                    alias.name not in WHITELISTED_LIBRARIES
-                    and alias.name != "pandas"
-                    or any(
-                        "." in alias.name
-                        and alias.name.rsplit(".", 1)[0] == whitelisted_lib
-                        for whitelisted_lib in WHITELISTED_LIBRARIES
-                    )
-                ):
+                if alias.name.split(".")[0] not in WHITELISTED_LIBRARIES:
                     raise BadImportError(alias.name)
 
         return False
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -323,9 +323,14 @@ class PandasAI:
                 if alias.name in WHITELISTED_OPTIONAL_LIBRARIES:
                     import_optional_dependency(alias.name)
                     continue
-                if not any(
-                    alias.name.startswith(whitelisted_lib)
-                    for whitelisted_lib in WHITELISTED_LIBRARIES
+                if (
+                    alias.name not in WHITELISTED_LIBRARIES
+                    and alias.name != "pandas"
+                    or any(
+                        "." in alias.name
+                        and alias.name.rsplit(".", 1)[0] == whitelisted_lib
+                        for whitelisted_lib in WHITELISTED_LIBRARIES
+                    )
                 ):
                     raise BadImportError(alias.name)
 
--- a/pandasai/constants.py
+++ b/pandasai/constants.py
@@ -10,7 +10,7 @@ END_CODE_TAG = "<endCode>"
 WHITELISTED_LIBRARIES = [
     "pandas",
     "numpy",
-    "matplotlib",
+    "matplotlib.pyplot",
 ]
 WHITELISTED_BUILTINS = [
     "abs",
--- a/pandasai/constants.py
+++ b/pandasai/constants.py
@@ -10,7 +10,7 @@ END_CODE_TAG = "<endCode>"
 WHITELISTED_LIBRARIES = [
     "pandas",
     "numpy",
-    "matplotlib.pyplot",
+    "matplotlib",
 ]
 WHITELISTED_BUILTINS = [
     "abs",
--- a/pandasai/constants.py
+++ b/pandasai/constants.py
@@ -8,7 +8,6 @@ It includes Start & End Code tags, Whitelisted Python Packages and While List Bu
 START_CODE_TAG = "<startCode>"
 END_CODE_TAG = "<endCode>"
 WHITELISTED_LIBRARIES = [
-    "pandas",
     "numpy",
     "matplotlib",
 ]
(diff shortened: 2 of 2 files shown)


pandasai/constants.py
```diff
@@ -8,8 +8,9 @@
 START_CODE_TAG = "<startCode>"
 END_CODE_TAG = "<endCode>"
 WHITELISTED_LIBRARIES = [
+    "pandas",
```
jonbiemond, line 11: Prior to https://github.com/gventuri/pandas-ai/commit/495f4c441ecfdd09982436994dc3dbc2ac2cfe92, the behaviour was that imports of `pandas` were removed from the run code. The logic as is would keep them in if added to WHITELISTED_LIBRARIES. Is that what we want?

jonbiemond (commented):
Would it be possible to include some tests? That way next time if my changes break things I'll be stopped 😅

pandasai/constants.py
```diff
@@ -8,8 +8,9 @@
...
     "numpy",
-    "matplotlib",
+    "matplotlib.pyplot",
```
jonbiemond, line 13: Alternatively we can check for the parent module in the list. That seems neater to me than updating the list to include submodules.

jonbiemond:
> Would it be possible to include some tests?

On second thought, I don't see why that's necessary. Nothing new being added here as of these commits.

pandasai/constants.py
sandiemann: yes, i remember but now we are raising the import error before generating the code atleast pandas can be on whitelist. maybe @gventuri can confirm?

pandasai/constants.py
sandiemann: yes, agree. it would be better. I'll add the check

pandasai/__init__.py
```diff
@@ -323,7 +323,10 @@ def _is_unsafe_import(self, node: ast.stmt) -> bool:
                 if alias.name in WHITELISTED_OPTIONAL_LIBRARIES:
                     import_optional_dependency(alias.name)
                     continue
-                if alias.name not in WHITELISTED_LIBRARIES:
+                if not any(
+                    alias.name.startswith(whitelisted_lib)
```
gventuri, line 327: This is scary, cause if a library whose name starts with another whitelisted library, we would allow it to be imported. This could cause some injections which we ideally want to avoid

sandiemann: yes, i though the same therefore explicitly defined the module in whitelist. This is extended based on the suggestion here: https://github.com/gventuri/pandas-ai/pull/211#discussion_r1214895423

pandasai/constants.py
sandiemann: yes I am aware but how do we surpass this since now we are throwing bad import error before generating & exec the code.

pandasai/__init__.py
jonbiemond: What if instead of `alias.name.startswith()` we look for `alias.name.split('.')[0]`?

pandasai/constants.py
jonbiemond: Currently when `_is_unsafe_import()` returns `True`, the unsafe library is removed from the generated code without raising an exception. When the `pandas` is imported, `_is_unsafe_import()` should return `True`. One possible implementation:
https://github.com/gventuri/pandas-ai/blob/55178f873d3b3db6e98d1726a42c7f5b8aa89bc8/pandasai/__init__.py#L321
```python
if alias.name in WHITELISTED_BUILTINS or alias.name == "pandas":
    return True
```

sandiemann:
I refactored the whitelist check to handle both the pandas & the submodule import

pandasai/__init__.py
```diff
@@ -323,9 +323,14 @@ def _is_unsafe_import(self, node: ast.stmt) -> bool:
                 if alias.name in WHITELISTED_OPTIONAL_LIBRARIES:
                     import_optional_dependency(alias.name)
                     continue
-                if not any(
-                    alias.name.startswith(whitelisted_lib)
-                    for whitelisted_lib in WHITELISTED_LIBRARIES
+                if (
+                    alias.name not in WHITELISTED_LIBRARIES
+                    and alias.name != "pandas"
```
jonbiemond, line 328: This is the same as adding `pandas` to `WHITELISTED_LIBRARIES`. So the line `import pandas as pd` persists and will be included in the run code. `_is_unsafe_import()` should return `True` for imports of pandas.

pandasai/__init__.py
```diff
@@ -323,9 +323,14 @@ def _is_unsafe_import(self, node: ast.stmt) -> bool:
...
+                    or any(
+                        "." in alias.name
+                        and alias.name.rsplit(".", 1)[0] == whitelisted_lib
+                        for whitelisted_lib in WHITELISTED_LIBRARIES
+                    )
```
jonbiemond, line 333: I believe we can one-line this condition using: `if alias.name.split(".")[0] not in WHITELISTED_LIBRARIES:`

jonbiemond (changes requested):

pandasai/__init__.py
jonbiemond: Here's a test:
```python
    def test_clean_code_remove_pandas(self, pandasai):
        pandas_code = """
import pandas as pd
print(pd.DataFrame())
"""
        pandasai._llm._output = pandas_code
        assert pandasai.run_code(pandas_code, pd.DataFrame()) == ""
        assert pandasai.last_run_code == "print(pd.DataFrame())"
```

sandiemann: ok yes, we should return. I'll refactor

pandasai/__init__.py
sandiemann: yes, no need for additional check for '.' since we only care about the first element

pandasai/__init__.py
jonbiemond: Sorry, my test was bad, here's a better one:
```python
    def test_clean_code_remove_pandas(self, pandasai):
        pandas_code = """
import pandas as pd
print(df.size)
"""
        pandasai._llm._output = pandas_code
        pandasai.run_code(pandas_code, pd.DataFrame())
        assert pandasai.last_run_code == "print(df.size)"
```

jonbiemond (approved):

pandasai/__init__.py
gventuri: @sandiemann @jonbiemond I'd definitely go for it! Can anyone see any cons?

sandiemann: @gventuri I've added the changes as suggested [here](https://github.com/gventuri/pandas-ai/pull/211/commits/95f2820afdfbfa4f4ef92205891a5e969927569b) Please review

PR #213, opened 2023-06-03
Return dataframe

### LINKED ISSUE #201: Dataframe is returned as a string
### 🐛 Describe the bug

When the model returns a dataframe, it is then converted to a string and cannot be used for further operations. We should make sure that when we return the dataframe we don't convert it to a string.

navneetdesai:
- [x] closes #201 

As discussed in #201, when the model returns a data frame, we should return it to the user as a data frame for further operations. 
Results that have a dataframe (with or without index) will be returned as dataframe instead of a string.

country             gdp  happiness_index
0   United States  19294482071552             6.94
1  United Kingdom   2891615567872             7.16
2          France   2411255037952             6.66
3         Germany   3435817336832             7.07
4           Italy   1745433788416             6.38
5           Spain   1181205135360             6.40
6          Canada   1607402389504             7.23
7       Australia   1490967855104             7.22

### DIFF
--- a/examples/from_dataframe.py
+++ b/examples/from_dataframe.py
@@ -5,7 +5,7 @@ import pandas as pd
 from pandasai import PandasAI
 from pandasai.llm.openai import OpenAI
 
-from .data.sample_dataframe import dataframe
+from data.sample_dataframe import dataframe
 
 df = pd.DataFrame(dataframe)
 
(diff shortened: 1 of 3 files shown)


gventuri:
Great job @navneetdesai, I'm just afraid it might be more involved trying to fix this. In fact, when the df is converted to a string some of the info might be lost and re-converting to df only works for some kind of df. Did you test your solution with bigger dataframes (i.e. more rows or columns?).

From my understanding, it's probably necessary to intervene before it's even converted to a string

MATRIX4284:
please refer to PR #223

navneetdesai:
Closing the PR for now.

PR #222, opened 2023-06-04
Update README.md

prtolem:
fixed examples


sandiemann (commented):
@prtolem we do not need to make specific run method call because we now allow run call via functional operator (__call__). This is introduced in version `v0.2.14` make sure you are on the latest version of pandasai

gventuri:
As @sandiemann mentioned, the `.run` method is called within `__call__`, so the example should work as it is, closing!

PR #225, opened 2023-06-05
fix: remove matplotlib from whitelisted libraries

### LINKED ISSUE #224: Matplotlib included in whitelisted libraries and environment variables
### 🐛 Describe the bug

Because `matplotlib` is included in `WHITELISTED_LIBRARIES` any import statements are persisted to the run code. It is also passed to the environment of the run code separately. I don't understand why exactly, but this results in any calls to `plt` to fail.

### Example
```python
df = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})

llm = OpenAI(api_token=<TOKEN>)
pandas_ai = PandasAI(llm, verbose=True, conversational=False)

pandas_ai(df, "Create a line chart of the dataframe.")
```
Returns nothing.

### Proposed Solution
* Rename `WHITELISTED_BUILTINS` to `ENVIRONMENT_LIBRARIES` and include `pandas` and `matplotlib`.

Note: Should `matplotlib` even be passed into the environment every time? It's only necessary when plotting.

jonbiemond:
`matplotlib` is added to the environment passed to `exec()`, it should not also be imported in the run code.
Remove `matplotlib` from `WHITELISTED_LIBRARIES`, so imports are removed from the generated code.
Add ENVIRONMENT_DEFAULTS variable to consolidate logic.

- [x] closes #224 
- [x] Integration test.

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -35,6 +35,7 @@ Example:
 import ast
 import io
 import re
+import sys
 from contextlib import redirect_stdout
 from typing import Optional
 
@@ -43,6 +44,7 @@ import matplotlib.pyplot as plt
 import pandas as pd
 
 from .constants import (
+    ENVIRONMENT_DEFAULTS,
     WHITELISTED_BUILTINS,
     WHITELISTED_LIBRARIES,
     WHITELISTED_OPTIONAL_LIBRARIES,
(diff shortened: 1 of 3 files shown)


pandasai/__init__.py
```diff
@@ -323,12 +325,15 @@ def _is_unsafe_import(self, node: ast.stmt) -> bool:
 
         if isinstance(node, (ast.Import, ast.ImportFrom)):
             for alias in node.names:
-                if alias.name in WHITELISTED_BUILTINS:
+                if (
+                    alias.name in WHITELISTED_BUILTINS
+                    or alias.name in ENVIRONMENT_DEFAULTS
```
jonbiemond, line 330: I'm not 100% satisfied with the implementation. But it works as a hotfix.

gventuri: I guess it's ok for now. We'll need to refactor it all at a later point!

PR #227, opened 2023-06-05
fix: replace path string with os.path.join

jonbiemond:
Replace hardcoded, Windows based, file path with a call to `os.path.join`. This way the path is valid on Unix machines as well.

- [x] Integration test on Windows machine
- [ ] Integration test on Unix machine

Note: I don't have easy access to a Unix environment. If someone wouldn't mind doublechecking that would be helpful.


jonbiemond:
Duplicate of #229

PR #234, opened 2023-06-06
feat: unique chart folder names

### LINKED ISSUE #230: Save charts to a unique folder
### 🚀 Feature request

Save charts to a unique folder per prompt.

              Agree that saving to a specific folder for each run, or let the user specify where the charts should be saved is important. This would make it easier to use pandasai inside of something like a fastapi, where the correct chart files can easily be sent through the API and allow for multiple api calls from different users at the same time.

_Originally posted by @mitokic in https://github.com/gventuri/pandas-ai/issues/212#issuecomment-1577204023_

jonbiemond:
Use a UUID to generate unique folders for each run of PandasAI to save charts to.
Add the UUID as a prompt id attribute to PandasAI, and open up with an interface in `last_prompt_id()`.
This allows projects built on top of PandasAI to easily retrieve charts generated.

- [x] closes #230 

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -36,6 +36,7 @@ import ast
 import io
 import re
 import sys
+import uuid
 from contextlib import redirect_stdout
 from typing import Optional, Union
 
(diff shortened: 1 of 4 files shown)


gventuri:
@jonbiemond this is super cool and goes in the direction of having several instances of PandasAI running and keeping the context!

pandasai/__init__.py
```diff
@@ -202,6 +205,8 @@ def run(
 
         self.log(f"Running PandasAI with {self._llm.type} LLM...")
 
+        self._prompt_id = str(uuid.uuid4())
```
gventuri, line 208: I would rather declare the process_id in the `__init__`. We could also have a combination of `prompt_id` and `process_id`. What do you think?

In case we want to keep also the prompt_id, let's print it so that in verbose mode we keep track of it!

PR #240, opened 2023-06-06
🐛 Add pd eval to cell value to correct bool evaluation

### LINKED ISSUE #209: 🐛 Error when trying to anonymize_dataframe_head a boolean type column
### 🐛 Describe the bug

When there is a boolean type column (not bool) in a dataframe, a problem arises when trying to apply the function anonymize_dataframe_head to that column. The error is not clear at all as it appears as:

> "Unfortunately, I was not able to answer your question, because of the following error:\n\nInvalid value 'True' for dtype boolean\n"

To replicate this error on v0.3.0, you can try:

```python
import os
from pandasai import PandasAI
import pandas as pd

llm = OpenAI(api_token=os.getenv('OPENAI_API_KEY'))
pandas_ai = PandasAI(llm)

df3 = pd.DataFrame(
    {'A': [1, 2, 3, 4, 5],
     'B': [True, True, True, True, True],}
)
df3.B = df3.B.astype('boolean')

ans = pandas_ai.run(
    df3,
    prompt='How many True values are in column B?'
    )
print(ans)

# Unfortunately, I was not able to answer your question, because of the following error:
# Invalid value 'True' for dtype boolean

```

Take into account that when the dtype is 'bool' there is no problem at all, so doing this, happens without problems:

```python
...
df3 = pd.DataFrame(
    {'A': [1, 2, 3, 4, 5],
     'B': [True, True, True, True, True],}
)
ans = pandas_ai.run(
    df3,
    prompt='How many True values are in column B?'
    )
print(ans)

# pandas as pd

# data = {'A': [5, 5, 4, 2, 3], 'B': [True, True, True, True, True]}
# df = pd.DataFrame(data)

# count_true = df['B'].sum()

# print(count_true)
# There are 5 True values in column B.
```

This is my pip list:
```bash
pip list | grep pandasai
# pandasai                  0.3.0
```

The error happens exactly on [this](https://github.com/gventuri/pandas-ai/blob/55178f873d3b3db6e98d1726a42c7f5b8aa89bc8/pandasai/helpers/anonymizer.py#L179) line:
```python
data_frame.iloc[random_row_index, col_idx] = cell_value
```
This is due to the function trying to put the value 'True' instead of True to a boolean type column. (bool admits 'True' and converts it).

I can work on a PR that solves this issue, it doesn't seem complicated at all, if the problem does deviate from what I think is happening.

alejohz:
- Closes #209 
- Tests added and passed.
- All code check passed.

### DIFF
--- a/pandasai/helpers/anonymizer.py
+++ b/pandasai/helpers/anonymizer.py
@@ -8,11 +8,11 @@ before sending it to the LLM (An External API). Only df.head() is sent to LLM AP
 import random
 import re
 import string
+
 import pandas as pd
 
 
 def is_valid_email(email: str) -> bool:
-
     """Check if the given email is valid based on regex pattern.
 
     Args:
@@ -26,7 +26,6 @@ def is_valid_email(email: str) -> bool:
 
 
 def is_valid_phone_number(phone_number: str) -> bool:
-
     """Check if the given phone number is valid based on regex pattern.
 
     Args:
@@ -40,7 +39,6 @@ def is_valid_phone_number(phone_number: str) -> bool:
 
 
 def is_valid_credit_card(credit_card_number: str) -> bool:
-
     """Check if the given credit card number is valid based on regex pattern.
 
     Args:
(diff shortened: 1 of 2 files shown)


gventuri:
@alejohz thanks a lot for the contribution! Great job!

PR #248, opened 2023-06-07
test: remove private method assertions

jonbiemond:
- Refactor tests closer to the interface.
- Remove unused imports and other linting.
- Resolve type hint errors.
- Add test for reading cache from file.

Motivation: https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice

### DIFF
--- a/tests/llms/test_base_llm.py
+++ b/tests/llms/test_base_llm.py
@@ -20,6 +20,8 @@ class TestBaseLLM:
         assert LLM()._polish_code(code) == "print('Hello World')"
         code = "`print('Hello World')`"
         assert LLM()._polish_code(code) == "print('Hello World')"
+        code = "print('Hello World')"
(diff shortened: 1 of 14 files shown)


tests/llms/test_base_llm.py
```diff
@@ -20,7 +20,6 @@ def test_polish_code(self):
         assert LLM()._polish_code(code) == "print('Hello World')"
         code = "`print('Hello World')`"
         assert LLM()._polish_code(code) == "print('Hello World')"
-        code = "print('Hello World')"
```
gventuri, line 23: I think this case should indeed be tested. Maybe we should add 
```
assert LLM()._polish_code(code) == "print('Hello World')"
```
instead of removing the line?

PR #250, opened 2023-06-07
👷 Update pre commit hook and project dev requirements for ruff

### LINKED ISSUE #247: Enh: Change pylint to ruff as default linter
Hey,

So ruff is 100 times faster than pylint and has in other projects been replaced with ease. Also, ruff can replace isort and pylint in the dev package requirements. This would let the dev dependencies like this:
```
[tool.poetry.group.dev.dependencies]
black = "^23.3.0"
pre-commit = "^3.2.2"
ruff = "^0.0.270"
pytest = "^7.3.1"
pytest-mock = "^3.10.0"
pytest-env = "^0.8.1"
click = "^8.1.3"
```

Additionally, the pre-commit only uses black and isort, we can also force a ruff revision, instead of isort for faster linting and checks.

Pylint vs Ruff is [widely mentioned](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-pylint) and Ruff developers stated that: Despite these differences, many users have successfully switched from Pylint to Ruff, especially those using Ruff alongside a [type checker](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-mypy-or-pyright-or-pyre), which can cover some of the functionality that Pylint provides.

I would be willing in working on a PR that addresses this. Please let me know if this is of interest.

alejohz:
- Closes #247

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,6 @@ jobs:
       - name: Install dependencies
         run: poetry install --all-extras
       - name: Lint with ruff
-        run: poetry run ruff check pandasai examples
+        run: poetry run ruff pandasai examples
       - name: Test with pytest
         run: poetry run pytest


alejohz:
We should run

```bash
poetry add ruff@^0.0.220
poetry remove isort
poetry remove pylint
```

The poetry.lock file is behind a keyring so i can't run these commands. With this, the CI checks should pass.

gventuri:
@alejohz updated the poetry.lock, but still the CI (which relies on pylint) fails.

jonbiemond:
The GitHub CI workflow should also be updated. https://github.com/gventuri/pandas-ai/blob/main/.github/workflows/ci.yml

gventuri:
Also, I think some rules will need to be changed to be in line with the default ones from pylint, or the linter will fail!

jonbiemond:
What if in a separate PR we resolve all `ruff` checks? And then we can rebase and merge this PR.

gventuri:
@jonbiemond could be an option, why not! Shouldn't be more, I think we can even just config a couple of settings (i.e. the max length of each line).

jonbiemond:
@alejohz If you've already resolved the `ruff` checks, then I think it's a good idea to include in this PR. Great work! 🎉

alejohz:
I added all the ruff changes and kept all queries equal on this [commit](https://github.com/alejohz/pandas-ai/commit/321776c2bf8811dba5e38380666f991a6f0774da). But, i don't see it here yet. Also, don't really know why checks are failing if ruff is installed [here](https://github.com/gventuri/pandas-ai/actions/runs/5202157449/jobs/9383358924?pr=250#step:5:116).

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: pip install poetry==1.4.2
       - name: Install dependencies
         run: poetry install --all-extras
-      - name: Lint with pylint
-        run: poetry run pylint pandasai examples
+      - name: Lint with ruff
+        run: bash scripts/lint.sh
```
jonbiemond, line 26: Maybe CI is failing because the script isn't running in the `poetry` environment. Why not just `poetry run rust check pandasai examples`?

gventuri: @alejohz did you have a chance to give it a try?

alejohz:
@gventuri @jonbiemond Thank you for the review! All checks passing and conlficts solved.

gventuri (approved):

gventuri:
@alejohz well done, merged! Congrats for the great job!

jonbiemond:
Looking at the GitHub action run times, it appears switching to `ruff` saves us ~15 seconds and is ~25% faster! 🙌🏻🎉

PR #272, opened 2023-06-11
fix: add whitelisted imports to exec environment

### LINKED ISSUE #253: Import statements in run code fail
### 🐛 Describe the bug

Import statements in the generated code that make it into the run code fail. 

### To Replicate
Run any code that imports an optional library through `pandasai.run_code()`.
For example:
```python
code = "import seaborn as sns\nprint(sns.__version__)"
pandasai.run_code(code, pd.DataFrame())
```
Returns: `""`

### Proposed Solution
I haven't figured out why but it seems that functions `exec` and `eval` aren't able to import libraries even though they're installed in the virtual environment. One solution would be to add any optional dependencies from the generated code to the environment passed to `exec`.

Related issues: #245, #238

jonbiemond:
- Consolidate WHITELISTED_OPTIONAL_LIBRARIES into WHITELISTED_LIBRARIES
- Add tests for different import scenarios
- Update clean_code() to remove all imports
- Refactor _is_unsafe_import() to _check_imports()

- [x] closes #253 

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -43,17 +43,13 @@ from contextlib import redirect_stdout
 from typing import List, Optional, Union
 
 import astor
-import matplotlib.pyplot as plt  # noqa: F401
 import pandas as pd
-
 from .constants import (
-    ENVIRONMENT_DEFAULTS,
     WHITELISTED_BUILTINS,
     WHITELISTED_LIBRARIES,
-    WHITELISTED_OPTIONAL_LIBRARIES,
 )
 from .exceptions import BadImportError, LLMNotFoundError
-from .helpers._optional import import_optional_dependency
+from .helpers._optional import import_dependency
 from .helpers.anonymizer import anonymize_dataframe_head
 from .helpers.cache import Cache
 from .helpers.notebook import Notebook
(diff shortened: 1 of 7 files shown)


tests/test_pandasai.py
```diff
@@ -265,15 +265,33 @@ def test_clean_code_remove_environment_defaults(self, pandasai):
...
 import numpy as np
-print(np.array([1, 2, 3]))
+np.array()
 """
         safe_code = safe_code.strip()
         pandasai._llm._output = safe_code
         assert pandasai.run_code(safe_code, pd.DataFrame()) == ""
-        assert pandasai.last_run_code == safe_code
+
+    def test_clean_code_whitelist_import_from(self, pandasai):
```
jonbiemond, line 278: This test currently fails!

pandasai/__init__.py
```diff
@@ -485,9 +519,10 @@ def run_code(
         )
 
         environment: dict = {
+            "pd": pd,
             **{
-                alias: sys.modules[library]
-                for library, alias in ENVIRONMENT_DEFAULTS.items()
+                lib["alias"]: getattr(import_dependency(lib["module"]), lib["name"])
```
jonbiemond, line 524: When adding a `ast.Import` to the environment, we don't need to get any attributes from the module. Just the module itself.

PR #290, opened 2023-06-15
fix: type hinting error

nautics889:
Fix type hint for `data_frame` parameter across methods of `PandasAI` class. According to [L275 of `__init__.py`](https://github.com/gventuri/pandas-ai/blob/main/pandasai/__init__.py#L275) the parameter can be a list of `DataFrame` objects.

- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -235,7 +235,7 @@ class PandasAI:
 
     def run(
         self,
-        data_frame: pd.DataFrame,
+        data_frame: Union[pd.DataFrame, List[pd.DataFrame]],
         prompt: str,
         is_conversational_answer: bool = None,
         show_code: bool = False,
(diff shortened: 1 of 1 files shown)


pandasai/__init__.py
```diff
@@ -235,7 +235,7 @@ def conversational_answer(self, question: str, answer: str) -> str:
 
     def run(
         self,
-        data_frame: pd.DataFrame,
+        data_frame: Union[pd.DataFrame, List[pd.DataFrame]],
```
jonbiemond, line 238: How do we feel about using the pipe operator instead of Union?
```python
from __future__ import annotations
...
data_frame: pd.DataFrame | list[pd.DataFrame]
```

gventuri: @jonbiemond IIRC `|` is only avaialble starting from python 3.10, so for the time being we should maybe use Union until we drop the support for older versions! What do you think?

nautics889: > How do we feel about using the pipe operator instead of Union?
> 
> ```python
> from __future__ import annotations
> ...
> data_frame: pd.DataFrame | list[pd.DataFrame]
> ```

I had the same thought at first, but as @gventuri has noted, it'd rather be better to use `Uniton` until we drop supports for <3.10 versions.

jonbiemond: That's where `from __future__ import annotations` comes in. 😀 It lets us use those `typing` features from `3.10` in `3.7` onwards! But really a minor thing and I'm sure there's a few places in the code we can replace imports of the `typing` module. Could be a little project for me one day.

PR #316, opened 2023-06-26
Feature/save charts with User Defined Path

### LINKED ISSUE #271: Save charts to folder specified by user
### 🚀 The feature

Allow the user to specify the folder they would like the charts to be saved in.

### Motivation, pitch

This way charts can be saved in a location more relevant to the user. For example if using PandasAI in multiple projects, each project's related charts can be saved in that project's folder. Also requested in https://github.com/gventuri/pandas-ai/issues/212#issuecomment-1577330648

### Alternatives

I see this as primarily a convenience feature. It does work as is.

### Implementation Proposal
One way this could be implemented is by passing the desired directory to the save_charts argument. 
For example: `pandas_ai = PandasAI(llm, save_charts='path/to/folder/')`
This does feel a little overloaded to me, but it keeps the interface neat.

### Additional context

_No response_

amjadraza:
PR related to #271 . Below is the example to use. 

```
import pandas as pd
import os
from data.sample_dataframe import dataframe

from pandasai import PandasAI

df = pd.DataFrame(dataframe)

llm = OpenAI()

user_defined_path = os.getcwd()
pandas_ai = PandasAI(llm, save_charts=True,
                     save_charts_path=user_defined_path,
                     verbose=True)
response = pandas_ai(
    df,
    "Plot the histogram of countries showing for each the gpd,"
    " using different colors for each bar",
)
# Output: check out $pwd/exports/charts/{hashid}/chart.png

```

- [ ] closes #271  


.gitignore
```diff
@@ -37,4 +37,6 @@ exports/
 *.log
 
 # vscode
-.vscode
\ No newline at end of file
+.vscode
+
+/exports/
```
jonbiemond, line 42: `exports/` can already be found on line 34.

pandasai/helpers/save_chart.py
```diff
@@ -64,9 +67,13 @@ def add_save_chart(code: str, folder_name: str, print_save_dir: bool = True) ->
 
     """
 
-    # define chart save directory
-    project_root = dirname(dirname(dirname(__file__)))
-    chart_save_dir = os.path.join(project_root, "exports", "charts", folder_name)
+    if save_charts_path is not None:
+        charts_root_dir = save_charts_path
```
jonbiemond, line 71: I think some validation here to make sure `save_charts_path` is a valid path would be good. What do you think?

pandasai/__init__.py
```diff
@@ -627,7 +629,10 @@ def run_code(
 
         # Add save chart code
         if self._save_charts:
-            code = add_save_chart(code, self._prompt_id, not self._verbose)
+
+            code = add_save_chart(code, self._prompt_id,
```
jonbiemond, line 633: Can you please double check you're running the code through `black`? I think this function call could be formatted to be a little more readable.

jonbiemond (changes requested):
Looks great! I reviewed the `save_charts` related features and had a couple thoughts.

pandasai/helpers/save_chart.py
amjadraza: I understand that it may be necessary to verify if the provided path is valid. However, I proceeded with the assumption that it is a string. Even if the path does not exist, the code will still generate the location.

One thing is notable, the previous implementation base path is the Path of File, Which seems not optimal.

pandasai/__init__.py
amjadraza: Will do it

.gitignore
amjadraza: Noted. However, it was not working for my Directory.

jonbiemond: That's surprising. It's formatted according to the docs: https://git-scm.com/docs/gitignore

amjadraza:
@jonbiemond addressed all your concerns. 

3. added `pandasai.__version__` : Reading version from `pyproject.toml` file

PR #336, opened 2023-07-02
Added auto-copy feature that copies code executed by pdai to clipboard.

### LINKED ISSUE #334: Auto Copy Code into Clipboard
### 🚀 The feature

The pandas_ai function will automatically copy the code that it generates into the user's clipboard if the argument copy_code is set to True.

### Motivation, pitch

Sometimes pandas ai produces an output very close to what is desired, but needs minor adjustments. Instead of prompt engineering, users can paste the generated code and tweak it to save time. 

### Proposed Usage

```py
# default
pandas_ai(
  df=df
  prompt=prompt)

# copy code
pandas_ai(
  df=df
  prompt=prompt
  copy_code=True)
```

yassinkortam:
- [x] closes #334 


gventuri:
@yassinkortam tests are failing!

yassinkortam:
To avoid using the pyperclip module as requested, I created my own copy and paste functions. They will only work on Unix / Linux based systems. There are a few edge cases I did not account for yet regarding quotation marks and so that will be fixed in a future commit.

gventuri:
@yassinkortam the test is failing. Don't know if it's because the test env doesn't have a "clipboard".

PR #339, opened 2023-07-03
tests: fix signature of a method in a fixture

nautics889:
Minor fix:
Update signature for `__call__()` method of `FakeOpenAI` mock according to it's base class.
Although it seems like it has no effect for now, because [the only place i found of usage](https://github.com/gventuri/pandas-ai/blob/main/pandasai/llm/langchain.py#L18) for that is just passing only `prompt` parameter. But i think it'd be better to keep method's original signature anyway, this basically seems to remove violation of Liskov Abstraction principle.

- [x] closes #xxxx (Replace xxxx with the GitHub issue number) _(No issue was created, minor fix)_

### DIFF
--- a/tests/llms/test_langchain_llm.py
+++ b/tests/llms/test_langchain_llm.py
@@ -16,7 +16,7 @@ class TestLangchainLLM:
         class FakeOpenAI(OpenAI):
             openai_api_key = "fake_key"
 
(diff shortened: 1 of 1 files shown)


gventuri:
@nautics889 nice catch, merging!

PR #354, opened 2023-07-12
fix langchain integration

hwchase17:
this allows ChatModels to be called as well!


gventuri:
@hwchase17 thanks a lot, it does the trick 🎉

PR #369, opened 2023-07-14
feat: passing log level in `log()` method (#368)

### LINKED ISSUE #368: Log level as a parameter for `log()` method of `PandasAI`
### 🚀 The feature

Currently implementation of the `log()` method looks like the following:
```
    def log(self, message: str):
        """Log a message"""
        self._logger.info(message)
        self._logs.append(message)
```
So, as you can see, all the logs are produced with `INFO` level (`20`).
I guess it would be great to add `level` as a parameter.

### Motivation, pitch

This  would let log messages with different levels (`logging.INFO`, `logging.ERROR`, etc.), perharps it should be extremely useful when catching some exceptions.

### Alternatives

_No response_

### Additional context

I propose to let `INFO` level be as a default value, so it wouldn't affect current code.

nautics889:
This commit adds possibility to specify log level when calling `.log()` method from `PandasAI` class.

The `INFO` level remains to be default value, so it won't affect any part of existing code.
___

* feat: update `log()` method of `PandasAI`, add passing log level as a parameter
* feat: add saving log level in logs container
* feat: update typing accordingly
* tests: update test for saving logs, add checking of logging for different levels

- [x] closes #368 

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -155,7 +155,7 @@ class PandasAI(Shortcuts):
     _start_time: float = 0
     _enable_logging: bool = True
     _logger: logging.Logger = None
-    _logs: List[str] = []
+    _logs: List[dict[str, str]] = []
     last_code_generated: Optional[str] = None
     last_code_executed: Optional[str] = None
     code_output: Optional[str] = None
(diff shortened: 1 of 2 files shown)


gventuri:
@nautics889 thanks a lot, nice addition! Just merged!

PR #378, opened 2023-07-19
feat : add callback functionality

gaurang98671:
* Added BaseCallback class with on_code abstract method
* Added StdoutCallback class which prints code returned by LLM
* Added test to check callback

All tests were successful
Hello @gventuri, please let me know if this can be merged. Thanks!


gaurang98671:
This will allows users to pass in their custom callbacks to intercept code returned by LLM

PR #417, opened 2023-07-30
chore: make packages pythonic

mspronesti:
This PR aims at fixing the structure of the packages, making them actually pythonic. 
Please have a close look: I'm only exposing the functions and classes that I believe the user should get access to. Make sure I haven't forgotten any before merging :)


gventuri:
@mspronesti great point, love it! Merged 🚀

PR #419, opened 2023-07-31
fix: environment for executing code

### LINKED ISSUE #400: Error executing code when plotting with Matplotlib
### 🐛 Describe the bug

I am trying to use PandasAI in conjunction with StarCoder to plot data using instructions, but it often fails when using matplotlib.

Here is an example, I wrote a simple script to load a dataset and prompt StarCoder using PandasAI to plot a pie chart:

```python
import pandas as pd
import matplotlib.pyplot as plt

from pandasai import PandasAI
from pandasai.llm.starcoder import Starcoder

SECRET_PATH = "secret.txt"
with open(SECRET_PATH, "r") as f:
    API_TOKEN = f.read()

llm = Starcoder(api_token=API_TOKEN)
pandas_ai = PandasAI(llm=llm, verbose=True)

DATA_PATH = "sales_data_sample.csv"

original_data = pd.read_csv(DATA_PATH, sep=",", encoding="ISO-8859-1")
original_data["ORDERDATE"] = pd.to_datetime(original_data["ORDERDATE"])
original_data = original_data.sort_values(by="ORDERDATE")

pandas_ai(original_data, "Plot a pie chart of sales by productline")
```

This generates correct code:
```python
df.groupby('PRODUCTLINE').sum()['SALES'].plot(kind='pie')
plt.show()
plt.close('all')
```

but fails with error:
```bash
<string>:1: FutureWarning: The default value of numeric_only in DataFrameGroupBy.sum is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.
2023-07-27 16:41:01 [WARNING] Error executing code (count: 0)
<string>:1: FutureWarning: The default value of numeric_only in DataFrameGroupBy.sum is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.
2023-07-27 16:41:04 [WARNING] Error executing code (count: 1)
<string>:1: FutureWarning: The default value of numeric_only in DataFrameGroupBy.sum is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.
2023-07-27 16:41:05 [WARNING] Error executing code (count: 2)
2023-07-27 16:41:06 [INFO] Answer: 
2023-07-27 16:41:06 [INFO] Executed in: 6.89627480506897s
Exception ignored in: <function Shelf.__del__ at 0x000001BF2AD1B910>
Traceback (most recent call last):
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\shelve.py", line 162, in __del__
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\shelve.py", line 144, in close
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\shelve.py", line 172, in sync
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\dbm\dumb.py", line 120, in _commit
TypeError: 'NoneType' object is not callable
Exception ignored in: <function _Database.close at 0x000001BF2C272F80>
Traceback (most recent call last):
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\dbm\dumb.py", line 274, in close
  File "C:\Users\asaju\AppData\Local\Programs\Python\Python310\lib\dbm\dumb.py", line 120, in _commit
TypeError: 'NoneType' object is not callable
```

Which is weird because if I run the generated code independently like so:
```python
import pandas as pd
import matplotlib.pyplot as plt

DATA_PATH = "sales_data_sample.csv"

original_data = pd.read_csv(DATA_PATH, sep=",", encoding="ISO-8859-1")
original_data["ORDERDATE"] = pd.to_datetime(original_data["ORDERDATE"])
original_data = original_data.sort_values(by="ORDERDATE")

original_data.groupby("PRODUCTLINE").sum()["SALES"].plot(kind="pie")
plt.show()
plt.close("all")
```

It does result in the correct pie chart:
image

Any ideas for a fix?

Here is the dataset used:
[sales_data_sample.csv](https://github.com/gventuri/pandas-ai/files/12184727/sales_data_sample.csv)

I am using:
- Python 3.10.6
- pandasai 0.6.12
- pandas 1.5.3
- matplotlib 3.7.2

### LINKED ISSUE #430: Try to import a package when `NameError` is raised during execution
### 🚀 The feature

When `NameError` is raised during the first code running, it possibly can indicate that there is usage of <name> in the code, but <name> is not imported. Perhaps there is should be a try for importing <name> as a package.

### Motivation, pitch

According to information provided in https://github.com/gventuri/pandas-ai/issues/400, the problem is related to cases when LLM provider "forgets" to add import of some library/package and uses it in a code.

The problem is also described in [this comment](https://github.com/gventuri/pandas-ai/issues/400#issuecomment-1661955799).

_Note:_ i've decided to create a new issue because the original one (https://github.com/gventuri/pandas-ai/issues/400) is essentially about `matplotlib.pyplot` itself.

nautics889:
This PR should resolve the problem with environment described in issue #400.

For now i've just added `matplotlib.pyplot` as "plt" in `_get_environment()` method. The package is present in dependency list, code from the issue runs without failures, it must works OK.

However it's still unclear for me, should we add the rest of popular packages, such as `numpy` as "np" and so on? Although `PandasAI` already has [`import_dependency()`](https://github.com/gventuri/pandas-ai/blob/main/pandasai/helpers/_optional.py#L45) util-function, and this logic in theory must have to avoid this problem, this  couldn't work because of (regarding to [the issue](https://github.com/gventuri/pandas-ai/issues/400)) HuggingFace just had returned code using `plt` variable without import statement — just something like `plt.show()`, and this caused `NameError`, since there was no such thing as `plt` in the context.

I've some idea for this cases, when API provider "forgets" to add import-statements. Like, for example, force adding `import <var_name>`  to the code when executing, if `NameError: name <var_name> is not defined` occured for example (btw this wouldn't work with aliases, but anyway can help with some other cases). But i'd like to hear opinion from others first.
@gventuri, @mspronesti welcome to discuss=)

___

* (fix): set `matplot.pyplot` package under "plt" alias in `_get_environment()`

___

- [x] closes #400
- [x] closes #430 

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -781,6 +781,11 @@ class PandasAI(Shortcuts):
                 )
             except Exception as new_exc:
                 exc = new_exc
+                self.log(
+                    f"Unable to fix `NameError`: an exception was raised: "
+                    f"{traceback.format_exc()}",
+                    level=logging.DEBUG,
+                )
 
         if not use_error_correction_framework:
             raise exc
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -48,7 +48,6 @@ import importlib.metadata
 
 __version__ = importlib.metadata.version(__package__ or __name__)
 import astor
-import matplotlib
 import pandas as pd
 from .constants import (
     WHITELISTED_BUILTINS,
@@ -643,7 +642,6 @@ class PandasAI(Shortcuts):
 
         return {
             "pd": pd,
-            "plt": matplotlib.pyplot,
             **{
                 lib["alias"]: getattr(import_dependency(lib["module"]), lib["name"])
                 if hasattr(import_dependency(lib["module"]), lib["name"])
@@ -773,7 +771,7 @@ class PandasAI(Shortcuts):
                 if search_name_res := re.search(name_ptrn, exc.args[0]):
                     name_to_be_imported = search_name_res.group(1)
 
-            if name_to_be_imported:
+            if name_to_be_imported and name_to_be_imported in WHITELISTED_LIBRARIES:
                 try:
                     package = __import__(name_to_be_imported)
                     environment[name_to_be_imported] = package
@@ -811,6 +809,13 @@ class PandasAI(Shortcuts):
                 f"[retry number: {retry_num + 1}]",
                 level=logging.WARNING,
             )
+        else:
+            self.log(
+                f"Unable to fix {exc.__class__.__name__} with error "
+                f"correction framework",
+                level=logging.ERROR,
+            )
+            raise exc
 
         return code
 
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -663,7 +663,7 @@ class PandasAI(Shortcuts):
             code (str): A python code
             e (Exception): An exception
             multiple (bool): A boolean to indicate if the code is for multiple
-            dataframes
+                             dataframes
 
         Returns (str): A python code
         """
@@ -703,6 +703,101 @@ class PandasAI(Shortcuts):
         code = self._llm.generate_code(error_correcting_instruction, "")
         if self.callback:
             self.callback.on_code(code)
+        return code
+
+    def _execute_catching_errors(
+        self, code: str, environment: dict
+    ) -> Optional[Exception]:
+        """
+        Perform execution of the code directly.
+
+        Call `exec()` for the given `code`, catch any non-base exceptions.
+
+        Args:
+            code (str): Python code
+            environment (dict): Context for the `exec()`
+
+        Returns (Optional[Exception]): Any exception raised during execution.
+                                       `None` if executed without exceptions.
+        """
+        try:
+            exec(code, environment)
+        except Exception as exc:
+            self.log("Error of executing code", level=logging.WARNING)
+            self.log(f"{traceback.format_exc()}", level=logging.DEBUG)
+
+            return exc
+
+    def handle_error(
+        self,
+        exc: Exception,
+        code: str,
+        environment: dict,
+        use_error_correction_framework: bool = True,
+        multiple: bool = False,
+    ) -> str:
+        """
+        Handle error occurred during first executing of code.
+
+        If `exc` is instance of `NameError`, try to import the name, extend
+        the context and then call `_execute_catching_errors()` again.
+        If OK, retunrs the code string; if failed, continuing handling.
+
+        If `use_error_correction_framework` is True, try to fix the error
+        via error correction framework withing `self._max_retries` number of
+        retries. If failed, raise an exception
+
+        Args:
+            exc (Exception): The caught exception.
+            code (str): Python code.
+            environment (dict): Context for the `exec()`
+            use_error_correction_framework (bool): If error correction
+                                                   framework should be used.
+                                                   Defaults to True.
+            multiple (bool): If the context has multiple dataframes.
+                             Defaults to False.
+
+        Raises:
+            Exception: Any exception which has been caught during
+                       the very first execution of the code.
+
+        Returns (str): Python code. Either an original or new, given by
+                       error correction framework.
+        """
+        if isinstance(exc, NameError):
+            try:
+                package = __import__(exc.name)
+                environment[exc.name] = package
+
+                caught_error = self._execute_catching_errors(code, environment)
+                if caught_error is None:
+                    return code
+
+            except ModuleNotFoundError:
+                self.log(
+                    f"Unable to fix `NameError`: package '{exc.name}' "
+                    f"could not be imported.",
+                    level=logging.DEBUG,
+                )
+            except Exception as new_exc:
+                exc = new_exc
+
+        if not use_error_correction_framework:
+            raise exc
+
+        for retry_num in range(self._max_retries):
+            code = self._retry_run_code(code, exc, multiple)
+            caught_error = self._execute_catching_errors(code, environment)
+
+            if caught_error is None:
+                break
+
+            self.log(
+                f"Failed to execute code with a correction framework "
+                f"[retry number: {retry_num + 1}]",
+                level=logging.WARNING,
+            )
+
         return code
 
     def run_code(
@@ -756,25 +851,15 @@ Code running:
 
         # Redirect standard output to a StringIO buffer
         with redirect_stdout(io.StringIO()) as output:
-            count = 0
-            while count < self._max_retries:
-                try:
-                    # Execute the code
-                    exec(code_to_run, environment)
-                    code = code_to_run
-                    break
-                except Exception as e:
-                    self.log(
-                        f"Error executing code (count: {count})", level=logging.WARNING
-                    )
-                    self.log(f"{traceback.format_exc()}", level=logging.DEBUG)
-
-                    if not use_error_correction_framework:
-                        raise e
-
-                    count += 1
-
-                    code_to_run = self._retry_run_code(code, e, multiple)
+            caught_error = self._execute_catching_errors(code_to_run, environment)
+            if caught_error:
+                code = self.handle_error(
+                    caught_error,
+                    code_to_run,
+                    environment,
+                    use_error_correction_framework=use_error_correction_framework,
+                    multiple=multiple,
+                )
 
         captured_output = output.getvalue().strip()
         if code.count("print(") > 1:
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -48,6 +48,7 @@ import importlib.metadata
 
 __version__ = importlib.metadata.version(__package__ or __name__)
 import astor
+import matplotlib
 import pandas as pd
 from .constants import (
     WHITELISTED_BUILTINS,
@@ -642,6 +643,7 @@ class PandasAI(Shortcuts):
 
         return {
             "pd": pd,
+            "plt": matplotlib.pyplot,
             **{
                 lib["alias"]: getattr(import_dependency(lib["module"]), lib["name"])
                 if hasattr(import_dependency(lib["module"]), lib["name"])
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -765,27 +765,36 @@ class PandasAI(Shortcuts):
                        error correction framework.
         """
         if isinstance(exc, NameError):
-            try:
-                package = __import__(exc.name)
-                environment[exc.name] = package
+            name_to_be_imported = None
+            if hasattr(exc, "name"):
+                name_to_be_imported = exc.name
+            elif exc.args and isinstance(exc.args[0], str):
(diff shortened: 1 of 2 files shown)


nautics889:
Thank for replying, @gventuri 
___

>it's more a game of figuring out how to import a given library if it hasn't been imported by the LLM for some reason

Yes, absolutely agree, that would be great if we can, for example, try to import a package if there were raised `NameError`. I'll upload some updates within this PR if you don't mind.
___
However there is still another point of the problem: how should we handle cases when LLM gives a code snippet **using a common alias**, like `np` for `numpy` or `plt` for `matplotlib.plt`, and without import statements (`import numpy as np` and so on). This is exactly the case appeared in https://github.com/gventuri/pandas-ai/issues/400.

nautics889:
recheck

nautics889:
Huh, as it turns out, `NameError` for python 3.9 does not have attribute `.name`, that's why tests failing only for 3.9. Already working on it...

nautics889:
@gventuri could you pls take a look?

___

Here's a brief description of what i've made within this PR:

* Split up code direct code execution and handling of errors to different methods. Drag them out from `run_code()`.
* Method `_execute_catching_errors()` just calls `exec(code, env)` wrapped by try-except, then returns exception if any errors appeared. Otherwise returns None.
* Method `handle_error()` is being called when at the very first execution of a code some exception was returned. It check if the error was an instance of `NameError`. If so, tries to fix that; otherwise continues troubleshooting. If `use_error_correction_framework` is True, tries to re-execute code, given by `_retry_run_code()`. Sure, it tries to do this `self._max_tries` times, or until `_execute_catching_errors()` returns None (meaning there was no exception during execution).

I've been trying to implement this small refactoring seamlessly, so essentially, there is no any change of original logic, except that a handling of `NameError` was added in addition to "error correction framework".

p. s. sorry for massive git diff

pandasai/__init__.py
```diff
@@ -642,6 +643,7 @@ def _get_environment(self) -> dict:
 
         return {
             "pd": pd,
+            "plt": matplotlib.pyplot,
```
gventuri, line 646: We don't need this, as we automatically import the relevant dependencies (with the right alias) with `import_dependency` a few lines below.

pandasai/__init__.py
```diff
@@ -661,7 +663,7 @@ def _retry_run_code(self, code: str, e: Exception, multiple: bool = False):
             code (str): A python code
             e (Exception): An exception
             multiple (bool): A boolean to indicate if the code is for multiple
-            dataframes
+                             dataframes
```
gventuri, line 666: This is probably a leftover?

tests/test_pandasai.py
```diff
@@ -146,6 +146,43 @@ def test_run_without_verbose(self, pandasai, llm):
...
+
+    def test_handle_error_name_error(self, pandasai):
+        code = "print(os)"
+        environment = {}
+
+        exc = pandasai._execute_catching_errors(code, environment)
+        assert code == pandasai.handle_error(
+            exc, code, environment, use_error_correction_framework=False, multiple=False
+        )
+        assert getattr(environment.get("os"), "__name__", None) == "os"
```
gventuri, line 184: Os is actually one of the libraries that we never want to allow, since it could potentially expose env variables and raise many more concerns 😅

pandasai/__init__.py
```diff
@@ -703,6 +705,115 @@ def _retry_run_code(self, code: str, e: Exception, multiple: bool = False):
...
+        """
+        try:
+            exec(code, environment)
+        except Exception as exc:
+            self.log("Error of executing code", level=logging.WARNING)
+            self.log(f"{traceback.format_exc()}", level=logging.DEBUG)
+
+            return exc
+
+    def handle_error(
```
gventuri, line 731: IMO here we should definitely check whether what we are trying to import is indeed a whitelisted library, or it's just too dangerous!

gventuri:
@nautics889 thanks a lot for the MR. Just left a couple of comments, mostly about security!

pandasai/__init__.py
nautics889: > we should definitely check whether what we are trying to import is indeed a whitelisted library
EXACTLY, had the same thought, i just needed to hear confirmation for this from you

tests/test_pandasai.py
nautics889: > Os is actually one of the libraries that we never want to allow, since it could potentially expose env variables and raise many more concerns 😅

Great point, thanks! Will try to find something else from stdlib...

pandasai/__init__.py
nautics889: > This is probably a leftover?

Not really... This comes mostly from a matter of taste whether to make visual indent for multilines description of arguments/possible exceptions/notes/return values. 
___
Let me show...

```python
def foobar(arg_one, arg_two):
    """
    ...

    Args:
        arg_one (str): Lorem ipsum dolor sit amet, consectetur adipiscing elit. 
        Vivamus bibendum mollis odio. Ut tellus nunc, tempor a elit sit amet, 
        fermentum eleifend justo. Nulla sit amet diam in erat finibus 
        arg_two (Exception): Lorem ipsum dolor sit amet, consectetur adipiscing elit. 
        Vivamus bibendum mollis odio. Ut tellus nunc, tempor a elit sit amet

    ...
    """
```
(no indend after new line for args verbal descriptions, kind of difficult to read as for me :) )

```python
def foobar(arg_one, arg_two):
    """
    ...

    Args:
        arg_one (str): Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            Vivamus bibendum mollis odio. Ut tellus nunc, tempor a elit sit amet,
            fermentum eleifend justo. Nulla sit amet diam in erat finibus
        arg_two (Exception): Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            Vivamus bibendum mollis odio. Ut tellus nunc, tempor a elit sit amet

    ...
    """
```
(indent is present for each new line in description, this should make docstring more readable as it separates arguments from each other)

___

References: [example of Google-style docstrings for python](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
image
___

HOWEVER, now i think it'd be better to make changes like that in a separate PR created specifically for docs/doc-style improvement. Of course, if you don't mind it.

pandasai/__init__.py
nautics889: > We don't need this, as we automatically import the relevant dependencies (with the right alias) with `import_dependency` a few lines below.

Roger that, with a next commit.

nautics889:
@gventuri thank you for the review!

I'm already working on a new commit; i've left some replies, will appreciate if you find time to take a look.

pandasai/__init__.py
gventuri: Got it, the few lines shown without expanding weren't providing me enough context! Makes a lot of sense then, thanks for clarifying!

pandasai/__init__.py
gventuri: Great, thanks a lot!

pandasai/__init__.py
gventuri: Sure, let's go for the allowing only whitelisted then! Thanks a lot!

tests/test_pandasai.py
gventuri: Great idea (and that makes me think that we might even need a "blacklist" list to prevent such dangerous libs from being exposed).

nautics889:
@gventuri 
I've uploaded an update. Added all the changes you've mentioned. 
____
However, there is an important point i've discovered during i was making the change. 

According to the logic of `run_code()` which we have now, when pandasai caught an exception during code running (and `use_error_correction_framework` is `True` of course), it tries to run code over and over again with the error correction framework, until in exceeds `_max_retries`. And in case when **the error correction framework can't manage to fix that** within `_max_retries` numbers of efforts, **non of exception will be raised**, silently, just the last form of code will be returned and processed further (in practice, sure, it definitely causes further errors). Is that logic correct and it's supposed to be like this?

So, i decided to raise the latest exception for such cases (sources [here](https://github.com/gventuri/pandas-ai/pull/419/commits/030b3aed5944f3b078f314a744674a7e7e0f09b6#diff-62b472d83558337a7726ff6758a8c431481784a74add715c96e2fed8bedbbd63R812)):
```python
        for retry_num in range(self._max_retries):
            code = self._retry_run_code(code, exc, multiple)
            caught_error = self._execute_catching_errors(code, environment)
            if caught_error is None:
                break
            self.log(
                f"Failed to execute code with a correction framework "
                f"[retry number: {retry_num + 1}]",
                level=logging.WARNING,
            )
        else:
            # `else` will only starts when `break` above didn't run in all the loop, i. e. ECF couldn't fix the error
            self.log(
                f"Unable to fix {exc.__class__.__name__} with error "
                f"correction framework",
                level=logging.ERROR,
            )
            raise exc

```
Also i added [tests for such cases](https://github.com/gventuri/pandas-ai/pull/419/commits/030b3aed5944f3b078f314a744674a7e7e0f09b6#diff-62b472d83558337a7726ff6758a8c431481784a74add715c96e2fed8bedbbd63R812).

___

Well, there is something else interesting i've encountered on. When i added the changes which i've described above, i found several tests cases failing. For example [this one](https://github.com/gventuri/pandas-ai/pull/419/commits/030b3aed5944f3b078f314a744674a7e7e0f09b6#diff-92144ec0277f612e0b022a692801edf93584cc8ecb551e0050411ba44207b3f5L389).
The reason for they were failing:
It's unable to create numpy's `array` with not passing any iterable object to it's `__init__()`.
_incorrect way:_
```python
In [1]: import numpy as np
In [2]: np.array()
Traceback (most recent call last):
  File "***\venvs\pandas-ai-venv\Lib\site-packages\IPython\core\interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-e4f3b47dc252>", line 1, in <module>
    np.array()
TypeError: array() missing required argument 'object' (pos 0)
```
_correct way:_
```python
In [1]: np.array([1, 2, 3])
Out[2]: array([1, 2, 3])
```
The reason why we didn't see those tests failed before:
As i've said
>**non of exception will be raised**, silently, just the last form of code will be returned and processed further
Considering this, tests were keeping silence, even when some exception actually had been raised when `exec()` code from LLM mocks.

___

Let me know if i'm wrong and exception suppressing should be returned back.

P. S. Apologize again for miscellaneous issues within a single PR and large git diff.

gventuri:
@nautics889 impressive job, thanks a lot for also fixing all the other edge cases you've highlighted! Having access to the last exception is a game-changer for debug! Merging!

PR #425, opened 2023-08-02
Added Poe-api as LLM reference

Iamprashanth-1:
- [x] closes #xxxx (Replace xxxx with the GitHub issue number)

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,42 @@ pandas_ai.generate_features(df)
 pandas_ai.plot_histogram(df, column="gdp")
 ```
 
+### POE-API As LLM Reference
+
+```
+To use Pandas-ai without using OPENAI API use Poe-api
+
+Get Your TOKEN from poe.com by just creating account don't need to add any billing
+
+To get Token click Inspect -> Application -> Cookies -> p-b value
+BOTS REFERENCE : 
+
+"chinchilla": "ChatGPT",
+  "a2": "Claude-instant",
+  "capybara": "Assistant",
+  "a2_100k": "Claude-instant-100k",
+  "llama_2_7b_chat": "Llama-2-7b",
+  "llama_2_13b_chat": "Llama-2-13b",
+  "a2_2": "Claude-2-100k",
+  "llama_2_70b_chat": "Llama-2-70b",
+  "agouti": "ChatGPT-16k",
+  "beaver": "GPT-4",
+  "vizcacha": "GPT-4-32k",
+  "acouchy": "Google-PaLM"
+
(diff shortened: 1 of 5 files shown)


Iamprashanth-1:
Adding Poe-api support because people don't have billing on their account or their API limit is exceeded they use this poe-api

@gventuri

mspronesti (commented):
Have you tried to run this ? I get a `AttributeError: 'NoneType' object has no attribute 'group'` (same specifying `formkey`, same installing `poe` with `-U`). 

Besides, this implementation contains quite a lot of leftovers from your copy-paste of the gpt4all PR and some encapsulation mistakes. If you have a fix for the first problem (which seems to be a known issue, see [here](https://github.com/ading2210/poe-api/issues/202)), I will be happy to leave a more detailed review.

Iamprashanth-1:
can you let me know which bot have you tried? mostly it works well with Assiatnt and ChatGpt

mspronesti:
I tried chinchilla, but the problem's not there. This simple snippet fails for me 
```py
import poe

token=...
formkey=...

client = poe.Client(token=token, formkey=formkey)
```

There are multiple issues reporting the same failed I mentioned in my previous comment. Seems to be something recurrent. Have a look [here](https://github.com/ading2210/poe-api/issues?q=is%3Aissue+AttributeError%3A+%27NoneType%27+object+has+no+attribute+%27group%27)

Iamprashanth-1:
Here Is the example:

from pandasai import PandasAI
from pandasai.llm.poe_api import POEAPI

llm = POEAPI(bot_name='chinchilla',token = '')
pandas_ai = PandasAI(llm,enable_cache=False , )
pandas_ai([df_tweets, df_tweet_users], "What are the top tweets?")

Just try this once:

don't need to provide formkey

mspronesti:
As explained above, it does not work for me neither with nor without the `formkey`. Also, to use your snippet I would need the dataframes :-) 
I used the `from_dataframe` example available in this repository replacing the `llm` and it yields the named error.

Iamprashanth-1:
but when I tried It's working fine. can you check your poe.com account whether promt is available in respective bot

mspronesti:
Of course. An invalid or missing token in poe api would raise the following error`RuntimeError: Invalid or missing token.` which is not the case for me. Anyway if it might be a problem with my setup, I let somebody else try this and give you feedback (even if that seems to be a very common problem among the poe api community).
If Gabriele wants to merge your PR, I will be happy to leave it a more thorough review.

Iamprashanth-1:
Can you update poe-api package to latest version and try once

PR #436, opened 2023-08-09
improved on _is_df_overwrite method to check all targets

### LINKED ISSUE #435: _is_df_overwrite function only checks the first target of assignments
### 🐛 Describe the bug

Hello,

I've been working with the `pandasai` package and noticed a potential oversight in the `_is_df_overwrite` function.

Here's the function for reference:

```python
def _is_df_overwrite(self, node: ast.stmt) -> bool:
    """
    Remove df declarations from the code to prevent malicious code execution.

    Args:
        node (object): ast.stmt

    Returns (bool):
    """

    return (
        isinstance(node, ast.Assign)
        and isinstance(node.targets[0], ast.Name)
        and re.match(r"df\d{0,2}$", node.targets[0].id)
    )
```

The issue arises here: 
```python
and isinstance(node.targets[0], ast.Name)
```

In the current implementation, only the first target is checked, potentially missing other assignments that could overwrite the DataFrame variable. I also experienced that with some prompts, where the output code still has a reassignment, it also contributes to higher API overhead. I would also suggest to  check for targets with different variable names than ```df```.  

I would suggest an implementation where all targets in an assignment statement are checked, ensuring that no potential DataFrame overwrites are missed. An example could be the following:
```python
def _is_df_overwrite(self, node: ast.stmt) -> bool:
    """
    Remove df declarations from the code to prevent malicious code execution.

    Args:
        node (object): ast.stmt

    Returns (bool):

    """
    if not isinstance(node, ast.Assign):
        return False

    for target in node.targets:
        if isinstance(target, ast.Name) and re.match(r"(df\d{0,2}|data)$", target.id):
            return True

    return False
```

I can create a PR for that, please let me know.

aymanehachcham:
- [x] closes #435.
- [x] Tests added and passed.

Fixed the _is_df_override method to check for all targets in an assignment statement, ensuring that no potential DataFrame overwrites are missed.

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -585,10 +585,12 @@ class PandasAI(Shortcuts):
 
         """
 
-        return (
-            isinstance(node, ast.Assign)
-            and isinstance(node.targets[0], ast.Name)
-            and re.match(r"df\d{0,2}$", node.targets[0].id)
+        if not isinstance(node, ast.Assign):
+            return False
+
+        return any(
+            isinstance(target, ast.Name) and re.match(r"(df\d{0,2}|data)$", target.id)
+            for target in node.targets
         )
 
     def _is_jailbreak(self, node: ast.stmt) -> bool:
--- a/pandasai/llm/__init__.py
+++ b/pandasai/llm/__init__.py
@@ -15,6 +15,5 @@ __all__ = [
     "GooglePalm",
     "GoogleVertexai",
     "LangchainLLM",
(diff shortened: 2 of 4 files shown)


aymanehachcham:
@gventuri This PR is ready to be merged, let me know if there is something else I can assist with.

gventuri:
@aymanehachcham thanks a lot for the fix. I've just added an additional test for this edge case.
Merging :)

PR #443, opened 2023-08-15
Updated smart dataframe to allow for sample_head in config.

yassinkortam:
# Added sample_head property to smart dataframe

Adding the ability to pass a custom sample head for smart dataframe or smart data lake. This allows users to anonymize their own data.

### Summary by CodeRabbit

**Release Notes**

- New Feature: Introduced a new optional parameter `_sample_head` to the `PandasAI` class constructor. This allows users to pass a sample dataframe that will be used for generating code.
- New Feature: Added a new field `sample_head` of type `pd.DataFrame` to the `Config` class in `df_config.py`.
- New Feature: Enhanced the `SmartDataframe` class with a new method `_load_config` for loading configuration for running queries.
- New Feature: Added a new attribute `_sample_head` to the `SmartDataframe` class, which stores a CSV string representation of the sample head of the dataframe.
- Test: Updated test fixtures and added new test methods in `tests/test_smartdataframe.py` to verify the getter and setter functionality of the `sample_head` attribute in the `SmartDataframe` class.

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -38,7 +38,6 @@ from typing import List, Optional, Union, Dict, Type
 import importlib.metadata
 
 import pandas as pd
-from .helpers.shortcuts import Shortcuts
 from .smart_dataframe import SmartDataframe
 from .smart_datalake import SmartDatalake
 from .prompts.base import Prompt
@@ -49,7 +48,7 @@ from .helpers.cache import Cache
 __version__ = importlib.metadata.version(__package__ or __name__)
 
 
-class PandasAI(Shortcuts):
+class PandasAI:
     """
     PandasAI is a wrapper around a LLM to make dataframes conversational.
 
@@ -98,7 +97,7 @@ class PandasAI(Shortcuts):
 
     """
 
-    _dl: SmartDatalake
+    _dl: SmartDatalake = None
     _config: Config
 
     def __init__(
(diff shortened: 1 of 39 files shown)


pandasai/__init__.py
```diff
@@ -147,6 +150,7 @@ def __init__(
             conversational=conversational,
             verbose=verbose,
             enforce_privacy=enforce_privacy,
+            sample_head=sample_head,
```
gventuri, line 153: I wouldn't add it here. This is just to ensure the backwards compatibility. We don't want new features to be supported here, or it will be tough to discontinue this at some point in the future! What do you think?

pandasai/helpers/df_config.py
```diff
@@ -13,6 +14,7 @@ class Config(BaseModel):
     save_logs: bool = True
     verbose: bool = False
     enforce_privacy: bool = False
+    sample_head: pd.DataFrame = None
```
gventuri, line 17: I would rather keep it as a dedicate param. Configs are most of the times settings that can be applied both to the datalake and dataframe.

pandasai/smart_dataframe/__init__.py
```diff
@@ -65,10 +65,14 @@ def __init__(
 
         self._load_df(df)
 
+        self._load_config(config)
```
gventuri, line 68: Wouldn't save _config in the smart df, since it always relies on the one that we store on the smart dl.

pandasai/smart_dataframe/__init__.py
```diff
@@ -90,6 +94,16 @@ def _load_df(self, df: DataFrameType):
         else:
             self._df = df
 
+    def _load_config(self, config: Config):
```
gventuri, line 97: I would remove this method too

pandasai/smart_dataframe/__init__.py
```diff
@@ -65,10 +65,14 @@ def __init__(
 
         self._load_df(df)
 
+        self._load_config(config)
+
         self._load_engine()
 
         self._dl = SmartDatalake([self], config=config, logger=logger)
 
+        self._sample_head = self._config.sample_head
```
gventuri, line 74: `self._sample_head = self.sample_head`

gventuri (commented):
Great job, just a couple of minor things. Also, would be great to write some tests!

PR #485, opened 2023-08-22
feat(llm): add support for TGI models (Llama2, CodeLlama, etc.)

### LINKED ISSUE #431: Support Llama v2 and Text generation inference
### 🚀 The feature

The feature consists in supporting the LLaMa v2 family of models (especially 70b, 70b-hf, 70b-chat and 70b-chat-hf) and potentially any model supported by text-generation-inference.

### Motivation, pitch

Hi @gventuri, 
I tested some of the llama v2 models (especially 70B and 70B chat) for pandas AI and they seem to be not bad at all. However, they are not available on Huggingface hosted inference API, hence private hosting is the only choice. The most performing way to do this is [text generation inference](https://github.com/huggingface/text-generation-inference) (TGI). 

However, once one writes this `HuggingfaceTextGenerationLLM`, any model supported by TGI can be easily supported in pandasAI, including falcon, starcoder and personal finetunes of those. Here's the proposal: how about I write a versatile class for TGI with a list of "officially" supported models for pandasAI (falcon, starcoder, the llama v2 family) and if the user passes a different one we throw a warning explaining we don't guarantee on the performance ? 

### Additional context

The same proposal above can be applied to the hosted inference API we are currently using for Falcon and Starcoder. In other words, instead of having one class per model we can just use `HuggingfaceLLM` (renaming it to `HuggingfaceEnpointLLM`, for instance) allowing any model but officially supporting only falcon and starcoder, as we are sure of their perfomances.

mspronesti:
Hi @gventuri,
this (draft) PR aims at supporting LLaMa-2 with [text generation inference](https://github.com/huggingface/text-generation-inference), addressing #431. 

This implementation supports both LLaMa-2 and LLaMa-2-Chat despite the first one seems (to me) to be much more stable, and thus defaulted. 

This PR is however a draft as I haven't ~~included any tests yet, nor~~ updated the docs. I'd like to let you and anybody else interested to play around with the model to finally assess whether it is worthy of being merged or not. Personally, I'm not in favor of merging models that don't perform "decently".

- [X] closes #431
- [X] Tests added and passed

### Summary by CodeRabbit

- New Feature: Introduced `HuggingFaceTextGen` class in the `pandasai.llm` module. This class allows users to generate text using HuggingFace models.
- New Feature: Added support for streaming text generation, allowing large-scale text generation tasks to be handled more efficiently.
- Test: Added comprehensive unit tests for the `HuggingFaceTextGen` class, ensuring its reliability and correctness.
- Documentation: Updated documentation with examples of how to use the new `HuggingFaceTextGen` class with a `SmartDataframe`.


gventuri:
@mspronesti looking into it, seems exciting 💯

gventuri:
@mspronesti cannot really test it, as it feels like it's not meant to work with M1/M2 chips :(

mspronesti:
@gventuri Unfortunately, I'm not aware of the support offered by tgi on M1/M2 chips. I recommend trying on a cloud instance if you have a subscription (e.g. on AWS using a p4d or a ml.g5.48xlarge EC2/SageMaker instance for LLaMa 70B and 70B Chat). Let me know if I can help in any way :)

gventuri:
@mspronesti tried it out and seems to work quite well on AWS! What do you think?

mspronesti:
@gventuri LLaMa-2-70B seems to work very well,  I have my reservations on LLaMa-2-70B-Chat though, which is a little unstable in my opinion. However, Meta has just released CodeLlaMa, which is about to be supported on text-generation-inference (it is already but there's a bug in the weights conversion). Once they fix it, we might also consider to support that model (with 0 lines of code needed wrt the current implementation of this PR) :)

zachschillaci27:
> @gventuri LLaMa-2-70B seems to work very well, I have my reservations on LLaMa-2-70B-Chat though. By the way, Meta has just released CodeLlaMa, which is about to be supported on text-generation-inference (it is already but there's a bug in the weights conversion). Once they fix it, we might also consider to support that model (with 0 lines of code needed wrt the current implementation of this PR) :)

Indeed, the coding capabilities of Llama-2 (even 70B) are rather bad, but Code Llama significantly improves things. It may also be worth trying Code Llama - Python, which is a Python-specialized variant of the model, see [here](https://arxiv.org/pdf/2308.12950.pdf). Here's an interesting excerpt taken from the paper:  "_Llama 2 was trained on 2T tokens, and training on only 500B of extra tokens from a code-heavy dataset results in massive performance gains on both HumanEval and MBPP, to the point that Llama 2 70B is roughly equivalent to Code Llama 7B on Python coding benchmarks._".

On a related note, I'm wondering if there's ever been a consideration to fine-tune a custom LLM for `pandasai`. With the release of Code Llama (7B, 13B and 34B + Python variants), I imagine one could train a powerful model using PEFT on a GPT-4 or human-generated dataset of Pandas examples. What do you think @gventuri?

pandasai/llm/llama2_text_gen.py
```diff
@@ -0,0 +1,89 @@
...
+            "do_sample": self.do_sample,
+            "seed": self.seed,
+        }
+
+    def call(self, instruction: Prompt, suffix: str = "") -> str:
+        prompt = instruction.to_string() + suffix
+        if self.use_chat_model:
+            # format prompt for LLaMa-2 Chat
+            prompt = f"[INST] <<SYS>>\n{prompt}\n<</SYS>>\n\n"
+
```
mspronesti, line 69: @zachschillaci27 yep, makes a lot of sense to me. Also, given the fact that this is the only part "llama2-specific" we could make this class a `HuggingfaceTextGeneration` LLM class to support any model and add some tutorials in the docs to explain how to specifically use LLaMa 2, CodeLlaMa, Starcoder etc.

mspronesti:
@gventuri The issue with the tensors conversion is now solved on tgi's side. I tested this implementation with CodeLlama-Python (34B, 13B, 7B) and I find them to be really decent! 
How about I rename this class `HuggingFaceTextGen` , removing the snippet highlighted above, and we add a few notebooks to showcase how to use it with Llama2, Llama2 chat, CodeLlama (and potentially, any model supported by TGI, such as StarCoder and Falcon)?
Does it make sense to you? :)

pandasai/llm/llama2_text_gen.py
gventuri: @zachschillaci27 that's a great idea, same for WizardCoder etc.. I'm playing around a little bit with it, will keep you posted, but it's definitely a thing that we'll work on ASAP!

gventuri:
@mspronesti sure, I agree about renaming it to `HuggingFaceTextGen` and adding some notebooks to showcase how to use it with the various models that we recommend/support!

mspronesti:
Ready to go :)

gventuri:
@mspronesti thanks a lot!! Looking forward to playing around with it in production! Merging 💯

PR #512, opened 2023-08-30
fix: langchain models throw LLMNotFoundError as a consequence of pyda…

### LINKED ISSUE #509: langchain models throw `LLMNotFoundError` as a consequence of pydantic's validator
### 🐛 Describe the bug

This issue is a generalization of #496, to better discuss how to tackle it.

In few: the validation performed at [this line](https://github.com/gventuri/pandas-ai/blob/main/pandasai/helpers/df_config.py#L30) will always fail for langchain's LLMs because [`load_llm`](https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L150-L169) - where the conversion is performed - is always called after creating the `Config` object [here](https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L75). 

Possible solutions:
- we perform the conversion at the level of the `Config` (simplest thing to do, in my opinion)
- we somehow load the llm before creating a new `Config` object. This only makes sense when `config` is a raw dictionary, and still forces the user to convert the llm on their own if they prefer to create a `Config` object to pass to `SmartDataFrame`'s constructor (which in fact expects to receive an object of type `pandasai.Config`)

gventuri:
…ntic's validator

- [x] closes #509

### Summary by CodeRabbit

- New Feature: Introduced support for `LangchainLLM` in the configuration, allowing users to utilize Langchain's language models.
- Refactor: Updated the `Config` class to validate the presence and type of the `llm` attribute using a root validator.
- Test: Enhanced test coverage for scenarios involving `LangchainLLM`, ensuring robustness of the new feature.
- Bug Fix: Added error handling in `LangchainLLM` initialization to handle missing dependencies and incorrect input types gracefully.

### DIFF
--- a/pandasai/config.py
+++ b/pandasai/config.py
@@ -2,6 +2,7 @@ import json
 from . import llm, middlewares, callbacks
 from .helpers.path import find_closest
 from .schemas.df_config import Config
+from .llm import LangchainLLM
 
 
 def load_config(override_config: Config = None):
@@ -31,6 +32,12 @@ def load_config(override_config: Config = None):
 
     if override_config:
         config.update(override_config)
+
+    if "llm" in config:
+        try:
+            config["llm"].is_pandasai_llm()
+        except AttributeError:
+            config["llm"] = LangchainLLM(config["llm"])
 
     config = Config(**config)
 
--- a/pandasai/llm/langchain.py
+++ b/pandasai/llm/langchain.py
@@ -11,7 +11,25 @@ class LangchainLLM(LLM):
     _langchain_llm = None
 
     def __init__(self, langchain_llm):
-        self._langchain_llm = langchain_llm
+        try:
+            from langchain.llms.base import BaseLLM
+
+            print("*" * 100)
+            print(langchain_llm)
+            print("*" * 100)
+
+            if not isinstance(langchain_llm, BaseLLM):
+                raise TypeError(
+                    "LangchainLLM must be initialized with a Langchain LLM."
+                )
+
+            self._langchain_llm = langchain_llm
+        except ImportError:
+            raise ImportError(
+                "Langchain is not installed. Please install it using "
+                "`pip install pandasai[langchain]` or `poetry install "
+                "pandasai[langchain]`."
+            )
 
     def call(self, instruction: Prompt, suffix: str = "") -> str:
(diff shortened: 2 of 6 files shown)


pandasai/config.py
```diff
@@ -32,6 +33,12 @@ def load_config(override_config: Config = None):
     if override_config:
         config.update(override_config)
 
+    if "llm" in config:
+        try:
+            config["llm"].is_pandasai_llm()
+        except AttributeError:
+            config["llm"] = LangchainLLM(config["llm"])
+
```
mspronesti, line 41: [Follow-up of what we are discussing in #509]
This should work fine, except that a user who wants to actually use pandasai's types and instantiate a `Config`  object to initialize their `SmartDataFrame` needs to manually do the conversion to `LangchainLLM`, as we are not doing that at the level of the `Config`. If we are good with this, I believe this is a good solution! :)

pandasai/llm/langchain.py
```diff
@@ -11,7 +11,25 @@ class LangchainLLM(LLM):
     _langchain_llm = None
 
     def __init__(self, langchain_llm):
-        self._langchain_llm = langchain_llm
+        try:
+            from langchain.llms.base import BaseLLM
+
+            print("*" * 100)
+            print(langchain_llm)
+            print("*" * 100)
```
mspronesti, line 19: Beware of this "debugging" leftovers 😆

pandasai/config.py
gventuri: I'd say at the moment there is not real reason to encourage users to instantiate their own `Config` (we are indeed trying to move configs to the json so that less and less config is required over time).

I'n not even in love with the idea of using pydantic validation to do transformations, that still seems a little bit of an anti-pattern to me. I'd say let's keep it this way for now (meaning that an advanced user that wants to instantiate a `Config`, will have no problem at figuring out how to instantiate a `LangChainLLM` wrapper).

If we figure out this is not the case, we might try to handle that edge case better. What do you think?

pandasai/llm/langchain.py
gventuri: Lol the rabbit had promptly complained about it too! 😛

PR #534, opened 2023-09-07
Version 1.2

gventuri:
### Summary by CodeRabbit

**Release Notes**

- New Feature: Introduced `DfValidator` for validating dataframes using Pydantic schemas.
- New Feature: Added flexibility to the configuration handling by allowing either a `Config` object or a dictionary in various classes and methods.
- New Feature: Added an optional `sample_head` parameter to the `SmartDataframe` class.
- New Feature: Updated chart saving functionality to allow custom directories.
- Deprecation: Marked `Starcoder` and `Falcon` classes as deprecated with warnings suggesting alternatives.
- Test: Added new test cases for dataframe validation and chart saving functionality.
- Documentation: Updated documentation with deprecation notices and minor changes.

### DIFF
--- a/docs/API/llms.md
+++ b/docs/API/llms.md
@@ -18,17 +18,21 @@ OpenAI API wrapper extended through BaseOpenAI class.
 options:
 show_root_heading: true
 
-### Starcoder
+### Starcoder (deprecated)
 
 Starcoder wrapper extended through Base HuggingFace Class
+
+- Note: Starcoder is deprecated and will be removed in future versions. Please use another LLM.
 
 ::: pandasai.llm.starcoder
 options:
 show_root_heading: true
 
-### Falcon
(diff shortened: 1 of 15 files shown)


mspronesti:
@gventuri Are we planning to also discontinue the `HuggingfaceLLM` class? After the suggestions we got in this thread #525, we could consider to keep that class, parametrizing the model name (to append it to the api) in order to allow users to use `starchat-beta`, `codellama-7b`, `codellama-13b`, `llama-2-7b` and any future fine-tune but without "promising" any performances, as we would with a dedicated class. 
Does it make sense? :)

gventuri:
@mspronesti I suppose we should basically change the implementation so that we go for the "official" way (not limited to 10 tokens with the "hacky" thing of running multiple calls to get the whole code generated

PR #562, opened 2023-09-15
feat: add `output_type` parameter (#519)

### LINKED ISSUE #519: Add Optional Parameter for Instruction Type in chat Method
### 🚀 The feature

Add a new optional parameter, `output_type`, to the `chat` method in the `generate_python_code.py` file. This parameter allows users to specify the type of instruction to be passed to the LLM. Depending on the value of `output_type`, the format of the instruction should be adjusted as follows:

- If `output_type` is "number," replace the instruction with:
  ```
  # - type (must be "number")
  # - value (must be a number)
  ```

- If `output_type` is "dataframe," replace the instruction with:
  ```
  # - type (must be "dataframe")
  # - value (must be a pandas dataframe)
  ```

- If `output_type` is "plot," replace the instruction with:
  ```
  # - type (must be "plot")
  # - value (must be a string containing the path of the plot image)
  ```

- If `output_type` is "string," replace the instruction with:
  ```
  # - type (must be "string")
  # - value (must be a conversational answer, as a string)
  ```

If `output_type` is not passed or is not one of the specified values, the instruction should remain as it is.

**Nice-to-Have:**
Implement logic to detect if the generated output differs from the required output based on the specified `output_type`. If a significant difference is detected, provide appropriate feedback or take corrective actions.

**Expected Output:**
Ensure that the `chat` method properly handles the `output_type` parameter and generates the correct instruction format based on the specified type.

### Motivation, pitch

By implementing this feature, users of your library will have greater flexibility in specifying the type of instruction they want to provide to the LLM, enhancing the library's usability and customization options.

### Alternatives

_No response_

### Additional context

_No response_

nautics889:
This PR supposed to implement a feature described in #519.

This PR adds a new parameter called `output_type` to be passed in `.chat()` methods. See [original requirements](https://github.com/gventuri/pandas-ai/issues/519#issue-1877521339) for details.

___

* (feat): update prompt template in `GeneratePythonCodePrompt`, add `output_type_hint` variable to be interpolated
* (feat): update `.chat()` method for `SmartDataframe` and `SmartDatalake`, add optional `output_type` parameter
* (feat): add `get_output_type_hint()` in `GeneratePythonCodePrompt` class
* (feat): add "output_type_hint" to `default_values` when forming prompt's template context
* (tests): update tests in `TestGeneratePythonCodePrompt`
* (tests): add tests for checking `output_type` interpotaion to a prompt

- [x] Closes #519
- [x] Add passing parameter `ouput_type` to `GeneratePythonCodePrompt`
- [x] Add a map for described output type names and according hints to be interpolated to prompt's template.
- [x] A new parameter covered with docstings.
- [x] Add logic for handling mismatch of output type passed by user and what the code generated by LLM actually returns.

### Summary by CodeRabbit

- New Feature: Introduced a flexible output type system in the `pandasai` package. Users can now specify the expected output type when interacting with the `SmartDataFrame` and `SmartDataLake` classes, improving the predictability of results.
- New Feature: Added validation for output types to ensure consistency and correctness of the returned data.
- Test: Expanded unit tests to cover new output type functionality and validation.
- Refactor: Simplified loop syntax and adjusted indentation in `huggingface_text_gen.py` for better readability.
- Chore: Updated function signatures and docstrings to reflect changes in output type handling.

### DIFF
--- a/pandasai/helpers/output_types/__init__.py
+++ b/pandasai/helpers/output_types/__init__.py
@@ -1,0 +1,45 @@
+from typing import Union
+
+from ._output_types import (
+    NumberOutputType,
+    DataFrameOutputType,
+    PlotOutputType,
+    StringOutputType,
+    DefaultOutputType,
+)
+
+output_types_map = {
+    "number": NumberOutputType,
+    "dataframe": DataFrameOutputType,
+    "plot": PlotOutputType,
+    "string": StringOutputType,
+}
+
+
+def output_type_factory(
+    output_type,
+) -> Union[
+    NumberOutputType,
+    DataFrameOutputType,
+    PlotOutputType,
+    StringOutputType,
+    DefaultOutputType,
+]:
+    """
+    Factory function to get appropriate instance for output type.
+
+    Uses `output_types_map` to determine the output type class.
+
+    Args:
+        output_type (str): A name of the output type.
+
+    Returns:
+        (Union[
+            NumberOutputType,
+            DataFrameOutputType,
+            PlotOutputType,
+            StringOutputType,
+            DefaultOutputType
+        ]): An instance of the output type.
+    """
+    return output_types_map.get(output_type, DefaultOutputType)()
--- a/pandasai/helpers/output_types/__init__.py
+++ b/pandasai/helpers/output_types/__init__.py
@@ -1,4 +1,5 @@
-from typing import Union
+import logging
+from typing import Union, Optional
 
 from ._output_types import (
     NumberOutputType,
@@ -7,6 +8,8 @@ from ._output_types import (
     StringOutputType,
     DefaultOutputType,
 )
+from .. import Logger
+
 
 output_types_map = {
     "number": NumberOutputType,
@@ -17,7 +20,7 @@ output_types_map = {
 
 
 def output_type_factory(
-    output_type,
+    output_type: str = None, logger: Optional[Logger] = None
 ) -> Union[
     NumberOutputType,
     DataFrameOutputType,
@@ -31,7 +34,12 @@ def output_type_factory(
     Uses `output_types_map` to determine the output type class.
 
     Args:
-        output_type (str): A name of the output type.
+        output_type (Optional[str]): A name of the output type.
+            Defaults to None, an instance of `DefaultOutputType` will be
+            returned.
+        logger (Optional[str]): If passed, collects logs about correctness
+            of the `output_type` argument and what kind of OutputType
+            is created.
 
     Returns:
         (Union[
@@ -42,4 +50,20 @@ def output_type_factory(
             DefaultOutputType
         ]): An instance of the output type.
     """
-    return output_types_map.get(output_type, DefaultOutputType)()
+    if output_type is not None and output_type not in output_types_map and logger:
+        possible_types_msg = ", ".join(f"'{type_}'" for type_ in output_types_map)
+        logger.log(
+            f"Unknown value for the parameter `output_type`: '{output_type}'."
+            f"Possible values are: {possible_types_msg} and None for default "
+            f"output type (miscellaneous).",
+            level=logging.WARNING,
+        )
+
+    output_type_helper = output_types_map.get(output_type, DefaultOutputType)()
+
+    if logger:
+        logger.log(
+            f"{output_type_helper.__class__} is going to be used.", level=logging.DEBUG
+        )
+
+    return output_type_helper
--- a/pandasai/helpers/output_types/_output_types.py
+++ b/pandasai/helpers/output_types/_output_types.py
@@ -1,0 +1,160 @@
+import re
+from decimal import Decimal
+from abc import abstractmethod, ABC
+from typing import Any, Iterable
+
+import pandas as pd
+import polars as pl
+
+
+class BaseOutputType(ABC):
+    @property
+    @abstractmethod
+    def template_hint(self) -> str:
+        ...
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        ...
+
+    def _validate_type(self, actual_type: str) -> bool:
+        if actual_type != self.name:
+            return False
+        return True
+
+    @abstractmethod
+    def _validate_value(self, actual_value):
+        ...
+
+    def validate(self, result: dict[str, Any]) -> tuple[bool, Iterable[str]]:
+        """
+        Validate 'type' and 'value' from the result dict.
+
+        Args:
+            result (dict[str, Any]): The result of code execution in
+                dict representation. Should have the following schema:
+                {
+                    "type": <output_type_name>,
+                    "value": <generated_value>
+                }
+
+        Returns:
+             (tuple(bool, Iterable(str)):
+                Boolean value whether the result matches output type
+                and collection of logs containing messages about
+                'type' or 'value' mismatches.
+        """
+        validation_logs = []
+        actual_type, actual_value = result.get("type"), result.get("value")
+
+        type_ok = self._validate_type(actual_type)
+        if not type_ok:
+            validation_logs.append(
+                f"The result dict contains inappropriate 'type'. "
+                f"Expected '{self.name}', actual '{actual_type}'."
+            )
+        value_ok = self._validate_value(actual_value)
+        if not value_ok:
+            validation_logs.append(
+                f"Actual value {repr(actual_value)} seems to be inappropriate "
+                f"for the type '{self.name}'."
+            )
+
+        return all((type_ok, value_ok)), validation_logs
+
+
+class NumberOutputType(BaseOutputType):
+    @property
+    def template_hint(self):
+        return """- type (must be "number")
+    - value (must be a number)"""
+
+    @property
+    def name(self):
+        return "number"
+
+    def _validate_value(self, actual_value: Any) -> bool:
+        if isinstance(actual_value, (int, float, Decimal)):
+            return True
+        return False
+
+
+class DataFrameOutputType(BaseOutputType):
+    @property
+    def template_hint(self):
+        return """- type (must be "dataframe")
+    - value (must be a pandas dataframe)"""
+
+    @property
+    def name(self):
+        return "dataframe"
+
+    def _validate_value(self, actual_value: Any) -> bool:
+        if isinstance(actual_value, (pd.DataFrame, pl.DataFrame)):
+            return True
+        return False
+
+
+class PlotOutputType(BaseOutputType):
+    @property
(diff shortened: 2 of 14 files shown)


pandasai/smart_datalake/__init__.py
```diff
@@ -280,10 +281,12 @@ def chat(self, query: str):
                 self.logger.log("Using cached response")
                 code = self._cache.get(self._get_cache_key())
             else:
+                prompt_cls = GeneratePythonCodePrompt
                 default_values = {
                     # TODO: find a better way to determine the engine,
                     "engine": self._dfs[0].engine,
                     "save_charts_path": self._config.save_charts_path.rstrip("/"),
+                    "output_type_hint": prompt_cls.get_output_type_hint(output_type),
```
gventuri, line 289: I suggest, instead of instantiating a static method here to pass `output_type` and then we assign the right path in the `__init__` method of the `GeneratePythonCodePrompt` class, like this:

```
def __init__(self, **kwargs):
        default_import = "import pandas as pd"
        engine_df_name = "pd.DataFrame"

        self.set_var("default_import", default_import)
        self.set_var("engine_df_name", engine_df_name)

        # assign it here
        self.set_var("output_type_hint", self.get_output_type_hint(kwargs["output_type"]))
```

nautics889: @gventuri 
Done. 
I thought to make exactly the same initially, but i was confusing a bit about `.set_var()` is called both from [`SmartDatalake._get_promt()`](https://github.com/gventuri/pandas-ai/blob/0eca724461843a272eeac929df042d4db52160be/pandasai/smart_datalake/__init__.py#L229) and [`GeneratePythonCodePrompt`'s `__init__()`](https://github.com/gventuri/pandas-ai/blob/0eca724461843a272eeac929df042d4db52160be/pandasai/prompts/generate_python_code.py#L78). And also that `__init__()` hadn't accepted keywords parameters (**kwargs) as well. So i've hesitated to add passing of `output_type`.

Fixed.

nautics889:
There are some tests failed also, will check it ASAP.

nautics889:
Tests are passed.

I'd like to ask to keep this in WIP state for a while to complete this option:
>Add logic for handling mismatch of output type passed by user and what the code generated by LLM actually returns

nautics889:
@gventuri looking forward for your feedback :)

___

Refactored the code a bit.

1. Create a class based approach for handling output types. 
2. Move all output types to [a separate module](https://github.com/gventuri/pandas-ai/pull/562/files#diff-44ba5efe405b02385a6b8b54ff85c6f34e441150977f2e3cbbc2275bc53a3e82), add [factory function](https://github.com/gventuri/pandas-ai/pull/562/files#diff-db43e372e2106ddfbdea93579b194d767fa86e7c0b87971aa3a9894198a1210fR19) for getting according output type object based on passed `output_type`.
3. Add `validate()` for those output types. Call `validate()` from `SmartDatalake` after getting a result dict.
4. Push warning messages to logs container if mismatch of the ouput type is detected.  
5. Add test method for checking if  log messages appears when the type returned by LLM in response is different from passed `output_type`.

___

I couldn't manage to create the correction logic yet, unfortunately; i'd like to add this within next PR's if it's possible. Sounds like it should be an addition to existing Error Correction Framework.

pandasai/helpers/output_types/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    Returns:
+        (Union[
+            NumberOutputType,
+            DataFrameOutputType,
+            PlotOutputType,
+            StringOutputType,
+            DefaultOutputType
+        ]): An instance of the output type.
+    """
+    return output_types_map.get(output_type, DefaultOutputType)()
```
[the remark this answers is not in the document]
gventuri: @nautics889 I think this kind of validation might be useful to prevent unexpected behaviors!

pandasai/helpers/output_types/_output_types.py
```diff
@@ -0,0 +1,160 @@
...
+    def template_hint(self):
+        return """- type (must be "dataframe")
+    - value (must be a pandas dataframe)"""
+
+    @property
+    def name(self):
+        return "dataframe"
+
+    def _validate_value(self, actual_value: Any) -> bool:
+        if isinstance(actual_value, (pd.DataFrame, pl.DataFrame)):
```
gventuri, line 94: I'd rather use df_config.df_type like this:

```
if df_type(actual_value) is not None:
```

gventuri: This also allows you not having to import polars and pandas!

tests/test_smartdataframe.py
```diff
@@ -251,9 +265,8 @@ def analyze_data(dfs: list[pd.DataFrame]) -> dict:
     2. Process: Manipulating data for analysis (grouping, filtering, aggregating, etc.)
     3. Analyze: Conducting the actual analysis (if the user asks to plot a chart save it to an image in exports/charts/temp_chart.png and do not show the chart.)
     4. Output: return a dictionary of:
-    - type (must be "number")
-    - value (must be a number)
-    Example output: { "type": "text", "value": "The average loan amount is $15,000." }
+    {output_type_hint}
+    Example output: {{ "type": "text", "value": "The average loan amount is $15,000." }}
```
gventuri, line 269: Maybe even the example output might change based on the type so that we don't risk hallucinating the LLM?

pandasai/helpers/output_types/_output_types.py
```diff
@@ -0,0 +1,160 @@
...
+    def _validate_type(self, actual_type: str) -> bool:
+        if actual_type != self.name:
+            return False
+        return True
+
+    @abstractmethod
+    def _validate_value(self, actual_value):
+        ...
+
+    def validate(self, result: dict[str, Any]) -> tuple[bool, Iterable[str]]:
```
gventuri, line 30: Well done, love it!

pandasai/smart_datalake/__init__.py
```diff
@@ -206,6 +207,7 @@ def _get_prompt(
         key: str,
         default_prompt: Type[Prompt],
         default_values: Optional[dict] = None,
+        output_type_hint: Optional[str] = None,
```
gventuri, line 210: I would pass it as part of default values in the `generate python code` prompt only, so that this is accessible both on the default prompt and in the custom prompts. This might not be needed in every prompt in the future.

gventuri (commented):
Great job, only a couple of minor notes!

pandasai/smart_datalake/__init__.py
nautics889: Sure, you're right, it makes the additional context to be in one place `default_values`. 
Apologize for spreading the setting logic like this, just we've got rly complicated implementation in `._get_prompt()` as for me. As i've mentioned: some variables're being set in `__init__()`, the rest are being set in `_get_prompt()` itself, when iterating over `default_values`. That was kind a tricky moment when i'd had output types mapping inside prompt class before.
___
Done

tests/test_smartdataframe.py
nautics889: Positive. Really great notice, this might have turned a dull confusion for the AI. I must just lost sight of that
___
Done

pandasai/helpers/output_types/_output_types.py
nautics889: Excellent, since this releases the module from having excessive imports of third party.
___
Done

pandasai/smart_datalake/__init__.py
gventuri: @nautics889 yeah, totally, it's super involved, we'll definitely need to simplify that a little bit in the future 😄

pandasai/helpers/output_types/__init__.py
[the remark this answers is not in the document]
nautics889: > I think this kind of validation might be useful to prevent unexpected behaviors!

This one was the most confusing for me.

Let me explain how i thought...

___

At first, it seemed appropriate to me to make an exception raising:
```python
def output_type_factory(output_type):
    if output_type is not None and output_type not in output_types_map:
        raise ValueError("Unknow 'output_type' has been passed. Possible are: ...")  # or custom exception
    return output_types_map.get(output_type, DefaultOutputType)()
```
In this way, user would be surely informed about pandasai can't give him a type which he/she expects.

However, the original requirements deny that:
> If output_type is not passed **or is not one of the specified values**, the instruction should remain as it is.

___

Well, then i try to log a warning message. Since we don't have a straight acces to pandasai logger, i've make the classic subpackage logging like this:
```python
import logging

...

logger = logging.getLogger(__name__)

...

def output_type_factory(output_type):
    if output_type is not None and output_type not in output_types_map:
        logger.warning("Unknow 'output_type' has been passed. Possible are: ...")
    return output_types_map.get(output_type, DefaultOutputType)()

```
Of course, this wouldn't give any to the stdout, since we [configure file handles with `.basicConfig`](https://github.com/gventuri/pandas-ai/blob/e06b3b977296aae5cd413041aa9217bb5c44b64d/pandasai/helpers/logger.py#L57) (i. e. only to a file). IDK, seems not too comprehensive for the end user. The logs are not appeared in `smart_df.logs` collection in this case ofc.
___
So, finally, i've add [an argument to pass for the logger object](https://github.com/gventuri/pandas-ai/pull/562/files#diff-db43e372e2106ddfbdea93579b194d767fa86e7c0b87971aa3a9894198a1210fR40); if passed some info about `output_type` is collected.
I wouldn't say it looks perfect for me, however we already have [something](https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L95) [like](https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L132) [this](https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L401), so i guess it's OK.

___

Btw, how about to consider making [`Logger` object](https://github.com/gventuri/pandas-ai/blob/main/pandasai/helpers/logger.py#L33) a [singleton](https://en.wikipedia.org/wiki/Singleton_pattern), then we could import/instantiate/pass/whatever in any of submodules and get the same object each time? Of course, this is for further feature requests and "to be discussed".

nautics889:
@gventuri 
Great notes above, thank you for the review!

gventuri:
@nautics889 thanks a lot for the great effort, looking forward to playing around with this feature (which IMHO is a super interesting one). I've merged on the `feature/1.3` and will release as part of the v1.3 so that in the meanwhile we can play around with it and check there are no regressions!

PR #563, opened 2023-09-15
Default `is_chat_model` to True in AzureOpenAI

mspronesti:
Hi @gventuri,
as OpenAI has marked the completion API as "legacy" (to be deprecated within 9 months) I have defaulted `is_chat_completion` to True in `AzureOpenAI`. Also, please notice that the newest subscription  in Azure OpenAI no longer allow to deploy completion-only models (such as davinci) and that the newest versions of turbo and gpt-4 (0613) no longer support the completion API.

### Summary by CodeRabbit

**Refactor:**
- Modified the `__init__` method in `azure_openai.py`. The `is_chat_model` parameter is now required and defaults to `True`, enhancing clarity and predictability when creating an instance of the class.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -37,7 +37,7 @@ nav:
       - Documents Building: building_docs.md
       - License: license.md
 extra:
-  version: "1.2"
+  version: "1.2.2"
 plugins:
   - search
   - mkdocstrings:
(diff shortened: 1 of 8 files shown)


gventuri:
Thanks a lot for the fix @mspronesti, merging! So basically in a few months we'll need to remove the completion function both from OpenAI and from AzureOpenAI, I suppose!

PR #585, opened 2023-09-21
fix[SqlAlchemy]: temporary downgrade sqlalchemy version to support al…

ArslanSaleem:
Downgrade SqlAlchemy version to support all connectors

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ openai = "^0.27.5"
 ipython = "^8.13.1"
 matplotlib = "^3.7.1"
 pydantic = "^1"
-sqlalchemy = "^2.0.19"
+sqlalchemy = "^1.4.49"
 duckdb = "^0.8.1"
 beautifulsoup4 = "^4.12.2"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -50,9 +50,11 @@ coverage = "^7.2.7"
 
 [tool.poetry.group.extras.dependencies]
 google-cloud-aiplatform = "^1.26.1"
+sqlalchemy-databricks = "^0.2.0"
+snowflake-sqlalchemy = "^1.5.0"
```
gventuri, line 54: @ArslanSaleem Why are these in the extras and how do the final user installs these?

mspronesti: Also, beware that `sqlalchemy-databricks` has been deprecated a while ago (see [here](https://github.com/crflynn/sqlalchemy-databricks)).

gventuri: I think that's the reason of the downgrade, but if an alternative is available we should definitely switch to that.

@ArslanSaleem here's the maintained version, I'll open a issue for that:
https://github.com/databricks/databricks-sql-python

PR #610, opened 2023-10-03
fix(llms): restore completion API for gpt-3.5-turbo-instruct

mspronesti:
Hi @gventuri,
this PR aims at restoring the `Completion` API as `gpt-3.5-turbo-instruct` is a completion model[^1]. Also, I restored `text-davinci-003` since it will be discontinued in 10 months but somebody might still be interested in using it for the time being. 

- [X] Tests added and passed if fixing a bug or adding a new feature

[^1]: despite OpenAI marking the `Completion` API as "Legacy", they have introduced  `gpt-3.5-turbo-instruct` as a replacement for all the other completions models

### Summary by CodeRabbit

- New Feature: Added support for the "gpt-3.5-turbo-instruct" model to the `OpenAI` class in the `pandasai/llm/openai.py` file.
- New Feature: Introduced a new list `_supported_completion_models` with two models: "text-davinci-003" and "gpt-3.5-turbo-instruct". This allows the `call` method to select the appropriate completion method based on the model used.
- Test: Added a new test case `test_call_supported_completion_model` in the `tests/llms/test_openai.py` file to ensure that the updated logic in the `call` method works as expected when using supported completion models.


gventuri:
Thanks a lot @mspronesti!

PR #633, opened 2023-10-10
feat: support more palm models

### LINKED ISSUE #632: Missing VertexAI models
### 🐛 Describe the bug

Currently, the wrapper for vertexai models has hardcoded value for model being code-bison-001 but google has many other variations of that model which will cause and error if used as model name for vertexai wrapper. 

```py
 def _generate_text(self, prompt: str) -> str:
        """
        Generates text for prompt.

        Args:
            prompt (str): A string representation of the prompt.

        Returns:
            str: LLM response.

        """
        self._validate()

        from vertexai.preview.language_models import (
            CodeGenerationModel,
            TextGenerationModel,
        )

        if self.model == "code-bison@001":
            code_generation = CodeGenerationModel.from_pretrained(self.model)

            completion = code_generation.predict(
                prefix=prompt,
                temperature=self.temperature,
                max_output_tokens=self.max_output_tokens,
            )```

### PR DESCRIPTION
**Pull request by gventuri**
- [x] closes #632

### Summary by CodeRabbit

**New Feature**
- Added support for multiple models in the `GoogleVertexAI` class. Users can now utilize different code and text models for their tasks.
- Introduced an `UnsupportedOpenAIModelError` exception. This error will be raised when a user attempts to use a model that is not supported.

**Test**
- Added a new test case `test_validate_with_invalid_model` to ensure the system correctly handles attempts to use unsupported models.

### DIFF
--- a/pandasai/llm/google_vertexai.py
+++ b/pandasai/llm/google_vertexai.py
@@ -12,7 +12,7 @@ Example:
 """
 from typing import Optional
 from .base import BaseGoogle
-from ..exceptions import UnsupportedOpenAIModelError
+from ..exceptions import UnsupportedModelError
 from ..helpers.optional import import_dependency
 
 
@@ -127,7 +127,7 @@ class GoogleVertexAI(BaseGoogle):
                 max_output_tokens=self.max_output_tokens,
             )
         else:
-            raise UnsupportedOpenAIModelError("Unsupported model")
+            raise UnsupportedModelError("Unsupported model")
 
         return str(completion)
 
--- a/pandasai/llm/google_vertexai.py
+++ b/pandasai/llm/google_vertexai.py
@@ -12,6 +12,7 @@ Example:
 """
 from typing import Optional
 from .base import BaseGoogle
+from ..exceptions import UnsupportedOpenAIModelError
 from ..helpers.optional import import_dependency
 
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pandasai/llm/google_vertexai.py [hunk 1]
```diff
@@ -115,6 +126,8 @@ def _generate_text(self, prompt: str) -> str:
                 top_k=self.top_k,
                 max_output_tokens=self.max_output_tokens,
             )
+        else:
+            raise UnsupportedOpenAIModelError("Unsupported model")
```
mspronesti, line 130: Shouldn't we generalize this exception? Or just throwing a `ValueError` would suffice, I suppose.

gventuri: Yep, correct! Done @mspronesti

arian81:
@gventuri I made improvement to the exception class to include the model name. I'm not sure the best way to approach to merge because it can't be added to this pull request. 
https://github.com/arian81/pandas-ai/tree/feat/support-more-models-palm

gventuri:
@arian81 thanks a lot! Let's do the following: we merge this and then you can submit another PR. Does it work for you?

gventuri:
> > @arian81 thanks a lot! Let's do the following: we merge this and then you can submit another PR. Does it work for you?
> 
> Sounds good to me.

Merged!

PR #634, opened 2023-10-10
chore: Improve unsupported model exceptions

### LINKED ISSUE #632: Missing VertexAI models

arian81:
- [x] closes #632

### Summary by CodeRabbit

**Refactor:**
- Renamed `UnsupportedOpenAIModelError` to `UnsupportedModelError` for better generality across different AI platforms. This change affects all modules and tests where this exception is used.

**New Feature:**
- Enhanced `UnsupportedModelError` to accept a `model_name` argument, providing more specific error messages when an unsupported model is encountered.

**Test:**
- Added a new test case `test_validate_with_invalid_model` in `test_google_vertexai.py` to validate the behavior when an invalid model is set.

**Bug Fix:**
- Fixed an issue in `azure_openai.py` where an unsupported model would not raise a clear exception. Now, `UnsupportedModelError` is raised with the specific `deployment_name` when encountered.

**Refactor:**
- Introduced a new exception `MissingModelError` for handling cases where the deployment name is not provided to Azure. This change is reflected in `azure_openai.py` and its corresponding tests.

### DIFF
--- a/pandasai/exceptions.py
+++ b/pandasai/exceptions.py
@@ -42,12 +42,12 @@ class MethodNotImplementedError(Exception):
     """
 
 
-class UnsupportedOpenAIModelError(Exception):
+class UnsupportedModelError(Exception):
     """
     Raised when an unsupported OpenAI model is used.
 
     Args:
-        Exception (Exception): UnsupportedOpenAIModelError
+        Exception (Exception): UnsupportedModelError
     """
 
 
--- a/pandasai/exceptions.py
+++ b/pandasai/exceptions.py
@@ -59,6 +59,15 @@ class UnsupportedModelError(Exception):
         )
 
 
+class MissingModelError(Exception):
+    """
+    Raised when deployment name is not passed to azure as it's a required paramter
+
+    Args:
+    Exception (Exception): MissingModelError
+    """
+
+
 class BadImportError(Exception):
     """
     Raised when a library not in the whitelist is imported.
--- a/pandasai/exceptions.py
+++ b/pandasai/exceptions.py
@@ -44,11 +44,18 @@ class MethodNotImplementedError(Exception):
 
 class UnsupportedModelError(Exception):
     """
-    Raised when an unsupported OpenAI model is used.
+    Raised when an unsupported model is used.
 
     Args:
-        Exception (Exception): UnsupportedModelError
+        model_name (str): The name of the unsupported model.
     """
+
+    def __init__(self, model_name):
+        self.model = model_name
+        super().__init__(
+            f"Unsupported model: The model '{model_name}' doesn't exist "
+            f"or is not supported yet."
+        )
 
 
 class BadImportError(Exception):
--- a/pandasai/llm/azure_openai.py
+++ b/pandasai/llm/azure_openai.py
@@ -17,7 +17,7 @@ from typing import Any, Dict, Optional
 import openai
 from ..helpers import load_dotenv
 
-from ..exceptions import APIKeyNotFoundError, UnsupportedModelError
+from ..exceptions import APIKeyNotFoundError, MissingModelError
 from ..prompts.base import AbstractPrompt
 from .base import BaseOpenAI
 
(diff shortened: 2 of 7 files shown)


arian81:
@gventuri I didn't create a whole new issue and just connected it to the old one since it's not that big of a change.

pandasai/llm/azure_openai.py
```diff
@@ -83,7 +83,8 @@ def __init__(
         openai.api_type = self.api_type
 
         if deployment_name is None:
-            raise UnsupportedModelError("Model deployment name is required.")
+            raise UnsupportedModelError(deployment_name)
```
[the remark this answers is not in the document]
gventuri: @arian81 I would raise another method or a generic exception with the message. It's more of a validation error (a param that is mandatory is not passed. What do you think?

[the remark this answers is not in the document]
arian81: I changed to a custom error for azure.

pandasai/llm/google_vertexai.py
```diff
@@ -127,7 +127,8 @@ def _generate_text(self, prompt: str) -> str:
                 max_output_tokens=self.max_output_tokens,
             )
         else:
-            raise UnsupportedModelError("Unsupported model")
+            raise UnsupportedModelError(self.model)
```
nautics889, line 130: @arian81 also please consider using the exactly name of a class instead of it's representation, make it to be something like this:
```python
raise UnsupportedModelError(self.model.__name__)
```
Just because we've got already the same like [here](https://github.com/gventuri/pandas-ai/blob/720811f9ebd5632a20aeff01ea7b73f125ba3a48/pandasai/prompts/file_based_prompt.py#L34).

Of course, this isn't important at all, it's on your choice, both of the ways look good as for me.

pandasai/exceptions.py
```diff
@@ -44,12 +44,29 @@ class MethodNotImplementedError(Exception):
...
+        self.model = model_name
+        super().__init__(
+            f"Unsupported model: The model '{model_name}' doesn't exist "
+            f"or is not supported yet."
+        )
+
+
+class MissingModelError(Exception):
+    """
+    Raised when deployment name is not passed to azure as it's a required paramter
```
nautics889, line 64:
```
    Raised when deployment name is not passed to Azure as it's a required parameter
```
typo

pandasai/llm/google_vertexai.py
arian81: I see that they used the other notation elsewhere but in the same logic code they're using self.model in the if statements. To me using other notation only for the exception looks inconsistent. Maybe there's a reasoning for using that notation in the other place, @gventuri can you elaborate on this ? 
```py
if self.model in self._supported_code_models:
            code_generation = CodeGenerationModel.from_pretrained(self.model)

            completion = code_generation.predict(
                prefix=prompt,
                temperature=self.temperature,
                max_output_tokens=self.max_output_tokens,
            )
        elif self.model in self._supported_text_models:
            text_generation = TextGenerationModel.from_pretrained(self.model)

            completion = text_generation.predict(
                prompt=prompt,
                temperature=self.temperature,
                top_p=self.top_p,
                top_k=self.top_k,
                max_output_tokens=self.max_output_tokens,
            )
        else:
            raise UnsupportedModelError(self.model)```

nautics889 (approved):
@arian81 

LGTM [👍](https://emojipedia.org/thumbs-up)

Enhances flexibility

pandasai/llm/google_vertexai.py
gventuri: @nautics889 thanks a lot for reporting, it's actually a very good point. I think in the other implementation, we were forced, since the template itself doesn't have a name, while the model does.
In this case, we want the specific model (i.e. code-bison) to be provided, as opposed to the LLM class.

PR #635, opened 2023-10-10
feat : Airtable connector Support

Tanmaypatil123:
- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)

Implemented Airtable connector with the help of [official docs](https://airtable.com/developers/web/api/list-records).
Required parameters are according to new [policy](https://support.airtable.com/docs/airtable-api-key-deprecation-notice) of airtable . Authorization `token`is used for authentication and `baseid` and `table_name` required to connect and load data.

### Summary by CodeRabbit

- New Feature: Introduced the `AirtableConnector` to the `pandasai` library, enabling users to connect to Airtable databases and retrieve data directly into a Pandas DataFrame. This feature enhances data accessibility and manipulation.
- New Feature: Added caching functionality to the `AirtableConnector` to improve performance by storing and reusing data.
- Documentation: Updated the `connectors.md` file to include instructions and examples on how to use the new `AirtableConnector`.
- Test: Added a comprehensive suite of unit tests for the `AirtableConnector` to ensure its functionality and reliability.
- New Feature: Enhanced error handling with the introduction of `InvalidRequestError` and `APIKeyNotFoundError` exceptions.

### DIFF
--- a/pandasai/connectors/airtable.py
+++ b/pandasai/connectors/airtable.py
@@ -11,12 +11,17 @@ from ..helpers.path import find_project_root
 import time
 import hashlib
 from ..exceptions import InvalidRequestError
+from functools import cache, cached_property
 
 
 class AirtableConnector(BaseConnector):
     """
     Airtable connector to retrieving record data.
     """
+
+    _rows_count: int = None
+    _columns_count: int = None
+    instance = None
 
     def __init__(
         self,
@@ -82,7 +87,7 @@ class AirtableConnector(BaseConnector):
             cache_dir = os.path.join(os.getcwd(), "cache")
         return os.path.join(cache_dir, f"{self._config.table}_data.parquet")
 
-    def _cached(self):
+    def _cached(self, include_additional_filters: bool = False):
         """
         Returns the cached Airtable data if it exists and
         is not older than the cache interval.
@@ -92,7 +97,7 @@ class AirtableConnector(BaseConnector):
                 it exists and is not older than the cache
                 interval, None otherwise.
         """
-        cache_path = self._get_cache_path()
+        cache_path = self._get_cache_path(include_additional_filters)
         if not os.path.exists(cache_path):
             return None
 
@@ -138,7 +143,15 @@ class AirtableConnector(BaseConnector):
         Returns:
             DataFrameType: The result of the connector.
         """
-        return self.fetch_data()
+        cached = self._cached() or self._cached(include_additional_filters=True)
+        if cached:
+            return pd.read_parquet(cached)
+
+        if isinstance(self.instance, pd.DataFrame):
+            return self.instance
+        else:
+            self.instance = self.fetch_data()
+        return self.instance
 
     def build_formula(self):
         """
@@ -159,7 +172,6 @@ class AirtableConnector(BaseConnector):
         """
         url = f"{self._root_url}{self._config.base_id}/{self._config.table}"
         params = {}
-
         if self._config.where:
             params["filterByFormula"] = self.build_formula()
         response = requests.get(
@@ -191,6 +203,7 @@ class AirtableConnector(BaseConnector):
         df = pd.DataFrame(records)
         return df
 
+    @cache
     def head(self):
         """
         Return the head of the table that
@@ -200,9 +213,14 @@ class AirtableConnector(BaseConnector):
             DatFrameType: The head of the data source
                  that the conector is connected to .
         """
-        return self.fetch_data().head()
+        # return self.fetch_data().head()
+        if isinstance(self.instance, pd.DataFrame):
+            return self.instance.head()
+        else:
+            self.instance = self.fetch_data()
+        return self.instance.head()
 
-    @property
+    @cached_property
     def rows_count(self):
         """
         Return the number of rows in the data source that the connector is
@@ -212,10 +230,13 @@ class AirtableConnector(BaseConnector):
             int: The number of rows in the data source that the connector is
             connected to.
         """
+        if self._rows_count is not None:
+            return self._rows_count
         data = self.execute()
-        return len(data)
+        self._rows_count = len(data)
+        return self._rows_count
 
-    @property
+    @cached_property
     def columns_count(self):
         """
         Return the number of columns in the data source that the connector is
@@ -225,8 +246,11 @@ class AirtableConnector(BaseConnector):
             int: The number of columns in the data source that the connector is
             connected to.
         """
+        if self._columns_count is not None:
+            return self._columns_count
         data = self.execute()
-        return len(data.columns)
+        self._columns_count = len(data.columns)
+        return self._columns_count
 
     @property
     def column_hash(self):
@@ -238,6 +262,8 @@ class AirtableConnector(BaseConnector):
             int: The hash code that is unique to the columns of the data source
             that the connector is connected to.
         """
-        data = self.execute()
-        columns_str = "|".join(data.columns)
+        if not isinstance(self.instance, pd.DataFrame):
+            self.instance = self.execute()
+        columns_str = "|".join(self.instance.columns)
+        columns_str += "WHERE" + self.build_formula()
         return hashlib.sha256(columns_str.encode("utf-8")).hexdigest()
--- a/pandasai/connectors/airtable.py
+++ b/pandasai/connectors/airtable.py
@@ -1,0 +1,92 @@
+"""
+Airtable connectors are used to connect airtable records.
+"""
+
+from .base import AirtableConnectorConfig, BaseConnector, BaseConnectorConfig
+from typing import Union, Optional
+import requests
+import pandas as pd
+
+
+class AirtableConnector(BaseConnector):
+    """
+    Airtable connector to retrieving record data.
+    """
+
+    def __init__(
+        self,
+        config: Optional[Union[AirtableConnectorConfig, dict]] = None,
+    ):
+        if isinstance(config, dict):
+            if config["token"] and config["baseID"] and config["table"]:
+                config = AirtableConnectorConfig(**config)
+
+        elif not config:
+            airtable_env_vars = {
+                "token": "AIRTABLE_AUTH_TOKEN",
+                "baseID": "AIRTABLE_BASE_ID",
+                "table": "AIRTABLE_TABLE_NAME",
+            }
+            config = AirtableConnectorConfig(
+                **self._populate_config_from_env(config, airtable_env_vars)
+            )
+
+        self._root_url: str = "https://api.airtable.com/v0/"
+
+        super().__init__(config)
+
+    def _init_connection(self, config: BaseConnectorConfig):
+        """
+        make connection to database
+        """
+        config = config.dict()
+        _session = requests.Session()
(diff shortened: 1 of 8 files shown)


pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class AirtableConnector(BaseConnector):
+    """
+    Airtable connector to retrieving record data.
+    """
+
+    def __init__(
+        self,
+        config: Optional[Union[AirtableConnectorConfig, dict]] = None,
```
gventuri, line 18: Let's add a cache mechanism similar to the other connectors!

gventuri: @Tanmaypatil123 check out yahoo finance connector for references

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+    ):
+        if isinstance(config, dict):
+            if config["token"] and config["baseID"] and config["table"]:
+                config = AirtableConnectorConfig(**config)
+
+        elif not config:
+            airtable_env_vars = {
+                "token": "AIRTABLE_AUTH_TOKEN",
+                "baseID": "AIRTABLE_BASE_ID",
```
gventuri, line 27: Could we make `baseID` it snake case (`base_id`)?

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+        self._root_url: str = "https://api.airtable.com/v0/"
+
+        super().__init__(config)
+
+    def _init_connection(self, config: BaseConnectorConfig):
+        """
+        make connection to database
+        """
+        config = config.dict()
+        _session = requests.Session()
```
gventuri, line 43: Why is `_session` underscored?

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+                ]
+                self._response = pd.DataFrame(records)
+            else:
+                raise ValueError(
+                    f"""Failed to connect to Airtable. 
+                    Status code: {_response.status_code}, 
+                    message: {_response.text}"""
+                )
+        return self._response
+
```
gventuri, line 82: Let's also add a `def fallback_name`, similarly to all the other connectors.

Example:
```
@property
    def fallback_name(self):
        """
        Return the fallback name of the connector.

        Returns:
            str: The fallback name of the connector.
        """
        return self._config.table
```

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+    def head(self):
+        """
+        Return the head of the table that
+          the connector is connected to.
+
+        Returns :
+            DatFrameType: The head of the data source
+                 that the conector is connected to .
+        """
+        return self._response.head()
```
gventuri, line 92: The head method should be independent. What happens if a `_response` hasn't been defined yet?

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,92 @@
...
+        self,
+        config: Optional[Union[AirtableConnectorConfig, dict]] = None,
+    ):
+        if isinstance(config, dict):
+            if config["token"] and config["baseID"] and config["table"]:
+                config = AirtableConnectorConfig(**config)
+
+        elif not config:
+            airtable_env_vars = {
+                "token": "AIRTABLE_AUTH_TOKEN",
```
gventuri, line 26: Let's rename to `api_key` to be consistent with the variable names in the LLMs.

pandasai/connectors/base.py
```diff
@@ -20,6 +20,16 @@ class BaseConnectorConfig(BaseModel):
     where: list[list[str]] = None
 
 
+class AirtableConnectorConfig(BaseConnectorConfig):
+    """
+    Connecter configuration for Airtable data.
+    """
+
+    token: str
```
gventuri, line 28: api_key

pandasai/connectors/base.py
```diff
@@ -20,6 +20,16 @@ class BaseConnectorConfig(BaseModel):
...
+    baseID: str
```
gventuri, line 29: base_id

gventuri (commented):
Let's try to increase the test coverage and let's add an example (from_airtable.py) and also let's add it on the connectors.md file

tests/connectors/test_airtable.py
```diff
@@ -0,0 +1,84 @@
...
+                },
+                {
+                    "id": "recsxnHUagIce7nB2",
+                    "createdTime": "2023-10-09T13:04:58.000Z",
+                    "fields": {"Name": "Campaign analysis", "Status": "To do"},
+                },
+            ],
+            "offset": "itrowYGFfoBEIob3C/recsxnHUagIce7nB2",
+        }
+        print(expected_data_json)
```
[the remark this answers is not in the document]
gventuri: @Tanmaypatil123 this is a leftover!

gventuri:
Additionally:
- the config should accept a param "where" similar to all the other connectors
- we should either handle pagination or see how many records we can get max (check out the maxRecords param)

tests/connectors/test_airtable.py
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        records = [
+            {"id": record["id"], **record["fields"]}
+            for record in expected_data_json["records"]
+        ]
+        expected_data = pd.DataFrame(records)
+        self.connector.head = Mock(return_value=expected_data)
+        head_data = self.connector.head()
+
```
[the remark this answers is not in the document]
gventuri: @Tanmaypatil123 mocking the method that you are testing is a bad practice, as it always returns the value you are forcing it to return.

gventuri:
Also, we should fallback to the env variables for base id and token, similarly to what we've done in the other connectors (i.e. mysql).

gventuri:
Btw did you try it? It doesn't seem to work for me.

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+        To prepare dataframe correctly.
+        """
+        records = [
+            {"id": record["id"], **record["fields"]} for record in data["records"]
+        ]
+
+        df = pd.DataFrame(records)
+        return df
+
+    def head(self):
```
ArslanSaleem, line 194: It should be cache using @cache for reference check sql.py

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        Return the head of the table that
+          the connector is connected to.
+
+        Returns :
+            DatFrameType: The head of the data source
+                 that the conector is connected to .
+        """
+        return self.fetch_data().head()
```
ArslanSaleem, line 203: here fetch only specific records lets say first 5 or random 5 rows instead of all.

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+    def rows_count(self):
+        """
+        Return the number of rows in the data source that the connector is
+        connected to.
+
+        Returns:
+            int: The number of rows in the data source that the connector is
+            connected to.
+        """
+        data = self.execute()
```
ArslanSaleem, line 215: Keep properties of rows_count and store it in instance so we don't have to call again and again

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+    def columns_count(self):
+        """
+        Return the number of columns in the data source that the connector is
+        connected to.
+
+        Returns:
+            int: The number of columns in the data source that the connector is
+            connected to.
+        """
+        data = self.execute()
```
ArslanSaleem, line 228: Same like rows count to not call again and again. For this use head function instead.

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+    def column_hash(self):
+        """
+        Return the hash code that is unique to the columns of the data source
+        that the connector is connected to.
+
+        Returns:
+            int: The hash code that is unique to the columns of the data source
+            that the connector is connected to.
+        """
+        data = self.execute()
```
ArslanSaleem, line 241: should also add where queries as well. Otherwise it will return wrong data from the cache.

pandasai/connectors/airtable.py
```diff
@@ -0,0 +1,243 @@
...
+        return self._config.table
+
+    def execute(self):
+        """
+        Execute the connector and return the result.
+
+        Returns:
+            DataFrameType: The result of the connector.
+        """
+        return self.fetch_data()
```
ArslanSaleem, line 141: You have cached it but never used it. You can for the cache like:

cached = self._cached() or self._cached(include_additional_filters=True)
        if cached:
            return pd.read_parquet(cached)

ArslanSaleem (changes requested):

pandasai/connectors/airtable.py
```diff
@@ -138,7 +143,15 @@ def execute(self):
         Returns:
             DataFrameType: The result of the connector.
         """
-        return self.fetch_data()
+        cached = self._cached() or self._cached(include_additional_filters=True)
+        if cached:
+            return pd.read_parquet(cached)
+
+        if isinstance(self.instance, pd.DataFrame):
```
ArslanSaleem, line 150: This self.instance is not need because we are already caching it

pandasai/connectors/airtable.py
```diff
@@ -11,13 +11,18 @@
...
 
 
 class AirtableConnector(BaseConnector):
     """
     Airtable connector to retrieving record data.
     """
 
+    _rows_count: int = None
+    _columns_count: int = None
+    instance = None
```
ArslanSaleem, line 24: This is extra because we already cache the dataframe

pandasai/connectors/airtable.py
```diff
@@ -200,9 +213,14 @@ def head(self):
             DatFrameType: The head of the data source
                  that the conector is connected to .
         """
-        return self.fetch_data().head()
+        # return self.fetch_data().head()
+        if isinstance(self.instance, pd.DataFrame):
```
ArslanSaleem, line 217: lets remove this instance for the head call api with params limit = 5

pandasai/connectors/airtable.py
```diff
@@ -147,60 +147,86 @@ def execute(self):
         if cached:
             return pd.read_parquet(cached)
 
-        if isinstance(self.instance, pd.DataFrame):
-            return self.instance
+        if isinstance(self._instance, pd.DataFrame):
+            return self._instance
         else:
-            self.instance = self.fetch_data()
-        return self.instance
```
ArslanSaleem: why this return is removed first time it will return None

pandasai/connectors/airtable.py
```diff
@@ -147,60 +147,86 @@ def execute(self):
...
+        data = self._preprocess(data=data)
         return data
 
-    def preprocess(self, data):
+    def _preprocess(self, data):
         """
         Preprocesses Json response data
         To prepare dataframe correctly.
         """
-        records = [
```
ArslanSaleem: Lets add this logic to _fetch_data function to get rid of three nested loops

PR #647, opened 2023-10-14
feat: Raise an exception when get error HTTP code (#646)

### LINKED ISSUE #646: Raising `LLMResponseHTTPError` when a remote provider responded with an error code
### 🚀 The feature

I've accidentally set wrong HF token when i was using `Starcoder` model. Then, i couldn't realize what the problem was that since [`query()` methods expects only a proper reponse](https://github.com/gventuri/pandas-ai/blob/f9facf383b6aff8e92065720a4719c5de11dc696/pandasai/llm/base.py#L335), sure, it had been failing when trying to get first item of the response JSON:
```python
        return response.json()[0]["generated_text"]
```
Even worse, the exception was caught [right here](https://github.com/gventuri/pandas-ai/blob/f9facf383b6aff8e92065720a4719c5de11dc696/pandasai/smart_datalake/__init__.py#L377), so, the only message i saw was:
```
0
```
_note:_ 0 is the index when getting first item in array from JSON-response i guess

### Motivation, pitch

I'd be good to raise error when LLM provider responses with error HTTP code, i believe.

### Alternatives

We can just log the error, but i doubt it would help to handle the problem.

### Additional context

_No response_

nautics889:
This PR aims to resolve the problem described in #646. 

Add:
- Raising `LLMResponseHTTPError` when an API for `HuggingFaceLLM` responses with 400+ code (e. g. invalid token, a remote server in under maintenance, etc.)
- According tests to check if the exception is raised

___

P. S. added the one only for `HuggingFaceLLM` class, because for others we have some different API calls (using `openai` SDK package for OpenAI for example)

___

* (feat): add `LLMResponseHTTPError` in exceptions.py
* (feat): raise `LLMResponseHTTPError` when HuggingFace LLM get HTTP code greater or equal to 400 (i. e. some error occurred)

- [x] Closes #646

### Summary by CodeRabbit

- New Feature: Introduced a new error handling mechanism for our software. Now, when our system interacts with external services and encounters an issue, it will provide more detailed feedback. This will help us diagnose and resolve issues more efficiently.
- Test: Added new tests to ensure this error handling works as expected. This will help maintain the reliability of our system and ensure a smooth user experience.

### DIFF
--- a/pandasai/exceptions.py
+++ b/pandasai/exceptions.py
@@ -68,6 +68,24 @@ class MissingModelError(Exception):
     """
 
 
+class LLMResponseHTTPError(Exception):
+    """
+    Raised when a remote LLM service responses with error HTTP code.
+
+    Args:
+        Exception (Exception): LLMResponseHTTPError
+    """
+
+    def __init__(self, status_code, error_msg=None):
+        self.status_code = status_code
+        self.error_msg = error_msg
+
+        super().__init__(
+            f"The remote server has responded with an error HTTP "
+            f"code: {status_code}; {error_msg or ''}"
+        )
+
+
 class BadImportError(Exception):
     """
     Raised when a library not in the whitelist is imported.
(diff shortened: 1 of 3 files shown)


gventuri:
@nautics889 thanks a lot for the improvement, merging!

PR #698, opened 2023-10-29
fix: initial directories creating

### LINKED ISSUE #654: Disable to create folders for exports and caches
### System Info

OS version: Ubuntu 20.04
Python version: Python 3.9.1
Pandasai version: 1.3.3

### 🐛 Describe the bug

I have tried to deploy the python script with pandasai to AWS lambda function. In AWS lambda function, it's not allowed to create folders like `exports`, `caches` (only under the `tmp` folders). So I tried to disable the `save_chart`, `enable_cache`, but it always creates such folders. I investigated the code and found this one.

https://github.com/gventuri/pandas-ai/blob/main/pandasai/smart_datalake/__init__.py#L77

```
    def initialize(self):
        """Initialize the SmartDatalake"""

        # Create exports/charts folder if it doesn't exist
        try:
            charts_dir = os.path.join((find_project_root()), "exports", "charts")
        except ValueError:
            charts_dir = os.path.join(os.getcwd(), "exports", "charts")
        os.makedirs(charts_dir, mode=0o777, exist_ok=True)

        # Create /cache folder if it doesn't exist
        try:
            cache_dir = os.path.join((find_project_root()), "cache")
        except ValueError:
            cache_dir = os.path.join(os.getcwd(), "cache")
        os.makedirs(cache_dir, mode=0o777, exist_ok=True)
```

It always run when create new SmartDataLake object, so it's nonsense.
I am curious, if the `save_chart`, `enable_cache` are disabled, it couldn't create such folders. Am I right?

Thank you.

nautics889:
This PR relates to #654.

1. Replace an order of calling of [`.initiliaze()`](https://github.com/gventuri/pandas-ai/blob/cbed878468bf8569951627799009814b12f8b5bb/pandasai/smart_datalake/__init__.py#L86) and [`._load_config()`](https://github.com/gventuri/pandas-ai/blob/cbed878468bf8569951627799009814b12f8b5bb/pandasai/smart_datalake/__init__.py#L88) methods. This allows the config object to be used in a scope of `.initialize()`, and 'config first' approach looks more correct. 
2. Add clauses in `.initialize()`: create directories for charts only if "save_charts" is passed as True. Same for cache as well.
3. Update tests accordingly.
___

* (fix): do not create directory for charts exporting if 'save_charts' is False
* (fix): do not create directory for cache files if 'enable_cache' is False as well
* (test): update tests accordingly

- [x] closes #654

## Summary by CodeRabbit

- New Feature: Enhanced the `SmartDatalake` class to conditionally create directories for saving charts and enabling cache based on user preferences. This change provides more flexibility and control to the user.
- Test: Added a new test case to verify the conditional creation of directories in the `SmartDatalake` class. This ensures the robustness of the new feature.

### DIFF
--- a/pandasai/smart_datalake/__init__.py
+++ b/pandasai/smart_datalake/__init__.py
@@ -83,9 +83,9 @@ class SmartDatalake:
             logger (Logger, optional): Logger to be used. Defaults to None.
         """
 
-        self.initialize()
-
         self._load_config(config)
+
+        self.initialize()
 
         if logger:
             self.logger = logger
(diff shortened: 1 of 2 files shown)


gventuri:
Thanks a lot for the fix @nautics889, merging!

PR #709, opened 2023-10-30
disable "unsafe" scipy in WHITELISTED_LIBRARIES

mtasic85:
Found exploit that leverages a flaw in the SciPy library, utilizing the imported ctypes module to execute arbitrary code.

PoC: https://github.com/tangledgroup/pandasai-sandbox-exploit/blob/2bdd4d7ebaea9c0d1633ee9cbccc256e1b5a2b33/exploit.py#L65

### DIFF
--- a/pandasai/constants.py
+++ b/pandasai/constants.py
@@ -84,6 +84,6 @@ WHITELISTED_LIBRARIES = [
     "json",
     "io",
     "base64",
-    "scipy",
+    # "scipy",
     "streamlit",
 ]


gventuri:
@mtasic85 thanks a lot for reporting. I'm wondering if we should put in place some mechanism to prevent this kind of attacks when the user asks to generate a specific code snippet. Because it could still happen someone needs to use scipy or find similar workarounds with other libraries.

Any ideas?

Issue #626, opened 2023-10-08
Return JSON in the response if requested by user


sahanxdissanayake:
### 🐛 Describe the bug

```python
import pandas as pd
from pandasai import SmartDatalake
from pandasai.llm import OpenAI

employees_data = {
    'EmployeeID': [1, 2, 3, 4, 5],
    'Name': ['John', 'Emma', 'Liam', 'Olivia', 'William'],
    'Department': ['HR', 'Sales', 'IT', 'Marketing', 'Finance']
}

salaries_data = {
    'EmployeeID': [1, 2, 3, 4, 5],
    'Salary': [5000, 6000, 4500, 9000, 5500]
}

employees_df = pd.DataFrame(employees_data)
salaries_df = pd.DataFrame(salaries_data)


llm = OpenAI(api_token="sk-")
dl = SmartDatalake([employees_df, salaries_df], config={"llm": llm, "verbose": True})
dl.chat("Who gets paid the least? return JSON in the exact format {person: name}")
```

error
```
[/usr/local/lib/python3.10/dist-packages/pandasai/smart_datalake/__init__.py](https://localhost:8080/#) in _add_result_to_memory(self, result)
    392             return
    393 
--> 394         if result["type"] == "string" or result["type"] == "number":
    395             self._memory.add(result["value"], False)
    396         elif result["type"] == "dataframe" or result["type"] == "plot":

KeyError: 'type'
```

llermaly:
Having a similar issue when trying to return multiple objects in one result. 

```
2023-10-08 02:04:10 [INFO] Answer: [{'type': 'number', 'value': {'F': 35, 'M': 32}}, {'type': 'plot', 'value': 'sex_counts.png'}]
```

Then 

```
  File "/opt/homebrew/lib/python3.11/site-packages/pandasai/smart_datalake/__init__.py", line 380, in chat
    self._add_result_to_memory(result)
  File "/opt/homebrew/lib/python3.11/site-packages/pandasai/smart_datalake/__init__.py", line 394, in _add_result_to_memory
    if result["type"] == "string" or result["type"] == "number":
       ~~~~~~^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

I'm trying to override the datalake with no luck

PR #730, opened 2023-11-03
fix: add checking type of 'result' (#626)

nautics889:
This PR adds an additional verification in `_add_result_to_memory()` of `SmartDatalake`.
Ensure that `result` is actually a dictionary.
Ensure that `result` contains `"type"` and `"value"` the items.

___

* (fix): add checking type of 'result' in '._add_result_to_memory()' method
* (fix): add checking that 'result' dict contains 'type' and 'value' the keys in '._add_result_to_memory()' method

- [x] Related to #626, but not closes it

## Summary by CodeRabbit

- **Refactor**
	- Enhanced input validation for the data addition process, improving error handling and logging.

### DIFF
--- a/pandasai/smart_datalake/__init__.py
+++ b/pandasai/smart_datalake/__init__.py
@@ -524,7 +524,20 @@ class SmartDatalake:
         Args:
             result (dict): The result to add to the memory
         """
-        if result is None:
+        if not isinstance(result, dict):
+            self.logger.log(
+                f"Inappropriate type of 'result' produced by generated code. "
+                f"Expected 'dict', got '{result.__class__.__name__}'",
+                level=logging.WARNING,
+            )
+            return
+
+        if "type" not in result or "value" not in result:
+            self.logger.log(
+                f"Both 'type' and 'value' items should be present in 'result' "
+                f"produced by generated code. Instead it contains the next "
+                f"content:\n{result}"
+            )
             return
 
         if result["type"] in ["string", "number"]:
--- a/pandasai/smart_datalake/__init__.py
+++ b/pandasai/smart_datalake/__init__.py
@@ -366,6 +366,8 @@ class SmartDatalake:
 
         self._memory.add(query, True)
 
+        result_is_valid = False
+
         try:
             output_type_helper = output_type_factory(output_type, logger=self.logger)
             viz_lib_helper = viz_lib_type_factory(self._viz_lib, logger=self.logger)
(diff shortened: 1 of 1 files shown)


pandasai/smart_datalake/__init__.py
```diff
@@ -524,7 +524,20 @@ def _add_result_to_memory(self, result: dict):
...
+                level=logging.WARNING,
+            )
+            return
+
+        if "type" not in result or "value" not in result:
+            self.logger.log(
+                f"Both 'type' and 'value' items should be present in 'result' "
+                f"produced by generated code. Instead it contains the next "
+                f"content:\n{result}"
+            )
```
gventuri, line 540: Great catch. I'm not entirely sure this is the place to do this check.

What if before we run `self._add_result_to_memory(result)` on line 508, we execute another method `self._validate_result` and in there we:
- check if both type and value are present
- check if the type is one of the ones accepted

If any of the conditions is not true, we raise an error.

What do you think? @nautics889

gventuri (changes requested):

pandasai/smart_datalake/__init__.py
nautics889: Thanks for the review, @gventuri.

___

You're right, there is [another validation](https://github.com/gventuri/pandas-ai/blob/3885c085ed0dae74878479e04007824f2e30679f/pandasai/smart_datalake/__init__.py#L469) a little bit before, therefore on [the 508'th line](https://github.com/gventuri/pandas-ai/blob/3885c085ed0dae74878479e04007824f2e30679f/pandasai/smart_datalake/__init__.py#L508) we likely can operate `validation_ok`. So, as turns out, the current PR brings an excessive validation for containing `"type"` and `"value"`. I've already noticed that before actually, but as seemed to me, it requires a bit of refactoring.

Working on it...

gventuri: Yes, great catch! Fortunately we are now working on simplifying the process a little bit. Too many things happen now and we need to improve the way we separate responsibilities.

nautics889:
Apologies for the delay.

@gventuri i'd be glad if you check the current update. At first i'd like to highlight that i was trying to **not change processing logic  (behaviour) at all**, since this is supposed to be a minor update. So that's why a call to `_add_result_to_memory()` still placed right after "Executed in" log. 

To summarize, the latest commit contains the next points:
1. Revert those excessive condition checks in `_add_result_to_memory()` have been added in the previous patch. 
2. Rename `validation_ok` → `result_is_valid` (seems less ambiguous to me)
3. Add pre-definition for `result_is_valid = False` [before the try-block](https://github.com/gventuri/pandas-ai/pull/730/files#diff-786be2fdc80d457ed86c7d296c417693a70a8b6ad3c645be4d24ba8e8f0c41c2R369). This allows us to operate `result_is_valid` later regardless whatever happened in try-exept block. So, if result is a dict having appropriate "type" and "value" items we call `_add_result_to_memory()`, in other cases (no matter it was an exception or result is None or some kind of mismatching else) `_add_result_to_memory()` won't be called.
4. As for [this part of diff](https://github.com/gventuri/pandas-ai/pull/730/files#diff-786be2fdc80d457ed86c7d296c417693a70a8b6ad3c645be4d24ba8e8f0c41c2L467-R487), this one is just to get rid from an excessive indent adding a complicated nested logic. So, basically the behaviour for this part left as it was.

___

Exuse me for the large diff.

gventuri:
@nautics889 thanks a lot for the refactor, as always 🚀! Merging it!

PR #750, opened 2023-11-13
fix: type hint for property `last_prompt_id`

nautics889:
Had already been [refactoring that a couple of months ago](https://github.com/gventuri/pandas-ai/pull/587/files#diff-786be2fdc80d457ed86c7d296c417693a70a8b6ad3c645be4d24ba8e8f0c41c2R52), but hadn't touched `@property` for the corresponding attribute in `SmartDatalake`, `SmartDataframe`.

It's better to let `last_prompt_id`'s return type be `uuid.UUID` rather than `str`. I understand the uuid indentifier is a text string itself, but python's `uuid.UUID` class is not a subclass of `str`.

___

* (fix): make type hint for the property `.last_prompt_id` to be `uuid.UUID` rather than `str`

___

## Summary by CodeRabbit

- **Refactor**
  - Updated the `last_prompt_id` property across various components to return a UUID instead of a string for enhanced uniqueness and consistency.

### DIFF
--- a/pandasai/__init__.py
+++ b/pandasai/__init__.py
@@ -35,6 +35,7 @@ Example:
 """
 import warnings
 from typing import List, Optional, Union, Dict, Type
+import uuid
 import importlib.metadata
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


gventuri:
@nautics889 thanks a lot for the fix, really appreciated!

PR #752, opened 2023-11-14
updating duckdb version

dioptre:
- [ ] Just adding outdated version of duck

Also breaking changes between minor versions until v1. so removed the ^ as it will create errors.


gventuri:
@dioptre check out, the build is failing: https://github.com/gventuri/pandas-ai/actions/runs/6869390905/job/18699701713?pr=752

PR #814, opened 2023-12-13
refactor: clean code in smart_dataframe and smart_datalake

gventuri:
## Summary by CodeRabbit

- **New Features**
  - Introduced class `Search` with methods for fetching and displaying results.
  - Added a search bar to the `Hero` component and a `Search` component to the `App` component.
  - Added styles for the search bar in `styles.css`.


pandasai/smart_datalake/__init__.py
```diff
@@ -30,167 +28,87 @@
...
-        else:
-            self._response_parser = ResponseParser(context)
-
-        if self._config.data_viz_library:
-            self._viz_lib = self._config.data_viz_library.value
-
-        self._conversation_id = uuid.uuid4()
-
-        self._instance = self.__class__.__name__
+        query_exec_tracker = QueryExecTracker(
```
ArslanSaleem, line 75: @gventuri  We can move this initialization to PipelineContext

gventuri: @ArslanSaleem great catch, but that's part of another refactor that I'd like to keep separated (which involves logging a new event on the query logger for each step of the pipeline. By that time, the whole QueryTracker logic will be part of the pipelines

pandasai/smart_dataframe/__init__.py
```diff
@@ -550,178 +180,55 @@ def last_error(self):
...
-        if isinstance(other, self.__class__) and (
-            self._core.has_connector and other._core.has_connector
-        ):
-            return self._core.connector.equals(other._core.connector)
-
-        return False
-
-    def is_connector(self):
-        return self._core.has_connector
+        return self.dataframe_proxy.connector.equals(other.dataframe_proxy.connector)
```
ArslanSaleem, line 214: @gventuri  .equals doens't exists for PandasConnector if gives error when using with direct_sql: true config.
"Unfortunately, I was not able to answer your question, because of the following error:\n\n'PandasConnector' object has no attribute 'equals'\n"

PR #897, opened 2024-01-25
[fix] logging chart saving only if code contains chart

### LINKED ISSUE #894: Minor fix in logger for chart
### System Info

version v1.5.17

### 🐛 Describe the bug

In method execute_code there is a call method add_save_chart when temporary chart is replaced by provided path.
Issue that it show message in the log always, even if I have nothing about chart in my query.
It's because no checking if we need to replace path or not:

```
        code = code.replace("temp_chart.png", save_charts_file.as_posix())
        logger.log(f"Saving charts to {save_charts_file}")

```
I guess if should be something like (but I'm not sure here):

```
   if "temp_chart.png" in code:
        code = code.replace("temp_chart.png", save_charts_file.as_posix())
        logger.log(f"Saving charts to {save_charts_file}")
```

Now we can have message in log only when we have somehing about chart.

Also maybe better to have "temp_chart.png" as const in some place.

There is another way to fix it - just replace message to "New place of chart is xxx"  because in fact we do not save chart in this code, we only replace path.

Lorenzobattistela:
- [x] Closes #894  (Replace xxxx with the GitHub issue number).

This is a simple fix where we prevent the logger to log "saving charts" when the code itself is not about charts.
We simply check if we have the default chart name in the code, and replace it by chart file name and log if it does.

I tried to create a const to save the default file name, but the sourcery complains, so we can keep simple and just test the string.

Possible reviewer @gventuri 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the chart saving functionality by ensuring temporary chart names are correctly replaced with designated file paths, enhancing the reliability of chart generation and saving.

### DIFF
--- a/pandasai/helpers/save_chart.py
+++ b/pandasai/helpers/save_chart.py
@@ -23,7 +23,6 @@ def add_save_chart(
         str: Code with line added.
 
     """
-
     save_charts_path = Path(save_charts_path_str) if save_charts_path_str else None
 
     if save_charts_path is not None:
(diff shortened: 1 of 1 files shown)


gventuri:
@Lorenzobattistela thanks a lot, merging :)

PR #907, opened 2024-01-28
feat: add support for modin

### LINKED ISSUE #657: Support for Modin
### 🚀 The feature

This is a feature proposal, more than a feature request: shall we support [`modin`](https://github.com/modin-project/modin) in pandas-ai? 

The implementation would require:
- an additional dependency
- optional logic to check if `modin` is installed, if marked as optional dependency
- replacing `pandas as pd` with `modin.pandas as pd` in the execution sandbox 

### Motivation, pitch

Quoting `modin`'s docs:
> Modin is a drop-in replacement for [pandas](https://github.com/pandas-dev/pandas). While pandas is single-threaded, Modin lets you instantly speed up your workflows by scaling pandas so it uses all of your cores. Modin works especially well on larger datasets, where pandas becomes painfully slow or runs [out of memory](https://modin.readthedocs.io/en/latest/getting_started/why_modin/out_of_core.html).
>
>By simply replacing the import statement, Modin offers users effortless speed and scale for their pandas workflows:
 ```py
import modin.pandas as pd
 ```

mspronesti:
Hi @gventuri, 
this draft PR finally adds support for `modin`! Please take your time to review it very carefully, as it introduces some core changes you should have a look at. Differently from my initial thought, this feature was not trivial at all. 

In addition, I have re-structured the dependencies to make `dev` and `docs` optional (before `poetry install` installed all of them).
I still have to update the docs and add some tests.

## Usage

```py
pandasai.set_pd_engine("modin") # or pandas

# use pandasai here!
```
- [X] Closes #657 
- [ ] Tests added and passed if fixing a bug or adding a new feature.

## Summary by CodeRabbit

- **New Features**
    - Enhanced the `smart_dataframe` module to support multiple data processing engines such as Pandas, Modin, and Polars.
    - Improved engine loading for different dataframes and connectors.
    - Extended the `chat` method to support various output types like pandas, modin, and polars dataframes.

- **Refactor**
    - Made changes to handle different dataframes and connectors efficiently.
    - Enhanced dataframe truncation capabilities for better performance.

- **Documentation**
    - Updated documentation to reflect the new engine support and dataframe handling capabilities.


mspronesti:
@gventuri You can start having a look if you like :) This PR is just missing a few tests.

pandasai/helpers/data_sampler.py
```diff
@@ -10,7 +10,8 @@
 import random
 
 import numpy as np
-import pandas as pd
+
+import pandasai.pandas as pd
```
gventuri, line 14: So if I understand correctly, to make it work with modin we need to import pandas like this. While if we import with `import pandas as pd` it will still work, but just ignore the logic to optionally import it with modin, right?

gventuri: @mspronesti while if one, for example, `import pandas as pd`, it will still work, right? What was the issue with `import modin.pandas as pd`?

mspronesti: @gventuri From the user perspective, pandasai is still supposed to be used with pandas or modin, i.e. one would just import one of the two in their snippet and use pandasai as they were before. 

However, internally this import is needed because `pandasai.pandas` contains the logic to select the engine, which is triggered whenenver `pandasai.set_pd_engine("modin|pandas")` is called.

gventuri: @mspronesti thanks a lot for clarifying, makes a lot of sense! I'll have a final review later today and merge it! Thanks a lot for the great improvement, super helpful for larger datasets!

Issue #956, opened 2024-02-24
Deprecation warning and failing at the function `predict` in LangChain with Ollama querying with chat class


rustem-dev:
### System Info

Ubuntu Server: 22.04.4 LTS
Python: 3.10.13
conda: 24.1.0
pandasai 1.5.19
ollama 0.1.27
langchain 0.0.199
langchain_community 0.0.24
LLM: ollama/phi

Warning while executing the script with locally running instance of  Ollama loaded into pandasai through langchain_community library. No script output is observed, killed with Ctrl + C. Seems like like the issue is with `chat` class which calls `predict` that is not supported in the langchain looking at warning message below.

```
from faker import Faker
import pandas as pd
from langchain_community.llms import Ollama
from pandasai import SmartDataframe
fake = Faker(locale='en_US')
langchain_llm = Ollama(model="phi")
# generate synthetic data and add into dataframe
def create_optics_data(rows_numbers):
 optics_data = []
 for i in range (1,rows_numbers):
     optics = {}
     optics['first_name'] = fake.first_name()
     optics['last_name'] = fake.last_name()
     optics_data.append(optics)
 return pd.DataFrame(optics_data)
df = pd.DataFrame(create_optics_data(10))
## wrapping into smartdataframe
df = SmartDataframe(df, config={"llm": langchain_llm})
df.invoke("how many names in dataframe?")
```

### 🐛 Describe the bug

```
/home/[myusername]/.local/lib/python3.10/site-packages/langchain_core/_api/deprecation.py:117: LangChainDeprecationWarning: The function `predict` was deprecated in LangChain 0.1.7 and will be removed in 0.2.0. Use invoke instead.
  warn_deprecated(
```

mspronesti:
Hi @rustem-dev. I've opened #957 to migrate the API. This will resolve the warning. However, I believe this is not likely to solve your problem with Ollama. Could you please try again with my fix and then report the error you get?

rustem-dev:
@mspronesti, if you referring to replacement of `predict` with `invoke` then I can confirm it works for me after the fix. Thank you!

mspronesti:
@rustem-dev Is Ollama also working properly now?

PR #957, opened 2024-02-24
chore(llms): upgrade to langchain API >= 0.1

mspronesti:
#956 

[OT] Nice to see no more "comments pollution" due to not-so-useful AIs!


gventuri:
Thanks a lot, merging!

[OT] yeah, so many comments, most of which garbage. And I realized at some point I was completely ignoring them!

PR #960, opened 2024-02-26
fix(llms): langchain response for chat models

mspronesti:
Langchain's integration is broken since a long time when used with chat models. This PR addresses that issue.


mspronesti:
Damn, I committed with the wrong user config. If you care about the commit history, I recommend amending the last commit to remove the co-author to pmprones 😆

PR #967, opened 2024-02-28
feat(workspace): adding workspace to chat

ArslanSaleem:


pandasai/workspace/__init__.py
```diff
@@ -0,0 +1,109 @@
+from typing import List, Optional
+
+import pandas as pd
```
mspronesti, line 3:
```suggestion
import pandasai.pandas as pd
```

pandasai/workspace/response_desrializer.py
```diff
@@ -0,0 +1,20 @@
+from typing import List
+
+import pandas as pd
```
mspronesti, line 3:
```suggestion
import pandasai.pandas as pd
```

mspronesti (changes requested):

PR #1008, opened 2024-03-08
feat(google_bigquery): load google big query from base64 string of cr…

ArslanSaleem:
…edentials

### DIFF
--- a/examples/from_googlebigquery.py
+++ b/examples/from_googlebigquery.py
@@ -1,25 +1,57 @@
+import json
 from pandasai import SmartDataframe
 
 # A license might be required for using Snowflake with PandasAI
 from pandasai.ee.connectors import GoogleBigQueryConnector
 from pandasai.llm import OpenAI
 
+import base64
+
+
 # ENV's
 # BIG_QUERY_DATABASE
 # KEYFILE_PATH
 # PROJECT_ID
 
+
+# EXAMPLE 1
 bigquery_connectors = GoogleBigQueryConnector(
     config={
-        "credentials_path": "credentials.json",
+        "credentials_path": "/Users/arslan/Downloads/loan-project.json",
         "database": "loan_payments",
         "table": "loan_payments",
-        "projectID": "project_id",
+        "projectID": "loan-project",
         "where": [["Gender", "=", "female"]],
     }
 )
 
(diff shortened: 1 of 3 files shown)


gventuri:
@ArslanSaleem there's a linting error!

examples/from_googlebigquery.py
```diff
@@ -1,25 +1,57 @@
...
+
+# Convert JSON data to a string
+json_string = json.dumps(json_data, indent=2)
+encoded_bytes = base64.b64encode(json_string.encode("utf-8"))
+
+
 bigquery_connectors = GoogleBigQueryConnector(
     config={
-        "credentials_path": "credentials.json",
+        "credentials_base64": encoded_bytes,
```
gventuri, line 46: @ArslanSaleem so we don't support the credentials_path anymore? Or we want to keep that as an option as well?

Also, is there any way that we can make it optional and automatically figure that out from the local environment if the user has been authenticated on the google cloud cli?

pandasai/ee/connectors/google_big_query.py
```diff
@@ -56,11 +64,15 @@ def _init_connection(self, config: GoogleBigQueryConnectorConfig):
...
-        )
+        if config.credentials_path:
+            self._engine = create_engine(
+                f"{config.dialect}://{config.projectID}/{config.database}",
+                credentials_path=config.credentials_path,
+            )
+        else:
+            self._engine = create_engine(
+                f"{config.dialect}://{config.projectID}/{config.database}?credentials_base64={config.credentials_base64}"
+            )
```
gventuri, line 75: Can we write a test case for this?

examples/from_googlebigquery.py
ArslanSaleem: @gventuri that i think is also possible but it need to use google.cloud lib should i do it?

pandasai/ee/connectors/google_big_query.py
ArslanSaleem: Done

examples/from_googlebigquery.py
gventuri: How involved would it be? Like you mean we need to create a custom version without sqlalchemy?

PR #1012, opened 2024-03-08
fix(pandas): modin support broke from pandasai>=2.0

### LINKED ISSUE #1011: pandasai 2.0 broke `modin` support
### 🐛 Describe the bug

Since #657, each module of pandasai  **must** import  `pandasai.pandas` as `pd` to make `pandasai` compatible with `modin`.

mspronesti:
- [X] Closes #1011

### DIFF
--- a/pandasai/agent/base.py
+++ b/pandasai/agent/base.py
@@ -3,7 +3,7 @@ import os
 import uuid
 from typing import List, Optional, Type, Union
 
-import pandas as pd
+import pandasai.pandas as pd
 
 from pandasai.pipelines.chat.chat_pipeline_input import (
     ChatPipelineInput,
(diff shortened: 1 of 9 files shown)


mspronesti:
It appears that many parts of `pandasai` broke when you merged the release branch (2.0) &ndash;  I recommend testing this again very thoroughly.

gventuri:
@mspronesti merging! Only a couple of considerations:
- do you think we can enforce this with some tests? I was trying to come up with an idea, got none
- I'm a little bit concerned this might happen again, as most of the times it's more automatic to `import pandas as pd` than `import pandasai.pandas as pd`, and I can't think of a way to enforce it

mspronesti:
@gventuri Probably rewriting the tests to run both for modin and pandas (i.e. setting `set_pd_engine("modin|pandas")`. I don't think this was the source of the problem though.

In fact, the current codebase contains leftovers and problems that arose from:
- Some incorrect conflict resolutions following the  PR containing 2.0
- Development in a text editor (such as VSCode, Atom, etc.) **which are not IDEs**. Using a real IDE, such as Pycharm, I see a lot of unused variables, and imports of classes that no longer exist (e.g. `CurrentCodePrompt`, `SimpleReasoningPrompt` [here](https://github.com/Sinaptik-AI/pandas-ai/blob/cf61cde92b80bce56fbcdea0da7cdab925df0e03/pandasai/prompts/direct_sql_prompt.py#L8-L12))

PR #992, opened 2024-03-04
fix: set Agent.last_code_executed correctly

chengwaikoo:
- [x] Fixes an issue where pipeline runs were not calling the `on_code_execution` callback, and so `Agent.last_code_executed` was not updated.
- [x] Tests added and passed if fixing a bug or adding a new feature.
- [x] All [code checks passed](https://github.com/gventuri/pandas-ai/blob/main/CONTRIBUTING.md#-testing).

PR #1015, opened 2024-03-09
fix: set Agent.last_code_executed correctly

chengwaikoo:
- [x] Fixes an issue where pipeline runs were not calling the `on_code_execution` callback, and so `Agent.last_code_executed` was not updated.

Repeat of #992 (can't be reopened) and targets the same issue as #1007 .

### DIFF
--- a/pandasai/agent/callbacks.py
+++ b/pandasai/agent/callbacks.py
@@ -30,7 +30,7 @@ class Callbacks:
         Args:
             code (str): A python code
         """
-        self.agent.last_code_executed = code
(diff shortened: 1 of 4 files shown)


gventuri:
Good catch @chengwaikoo! I was under the impression we were only executing the code only and that the code had already been cleaned by then.

I've created a new task to split the logic on the CodeExecution in 2: CodeCleaning and CodeExecution, so that we can keep the 2 concerns separated: https://github.com/Sinaptik-AI/pandas-ai/issues/1017

Merging :)

PR #1021, opened 2024-03-11
feat: Add Bedrock Claude 3 support

goriri:
- [ ✓] Tests added and passed if fixing a bug or adding a new feature.
- [ ✓] All [code checks passed](https://github.com/gventuri/pandas-ai/blob/main/CONTRIBUTING.md#-testing).


gventuri:
@goriri thanks a lot for the great PR!

Only a couple of things:
* It's failing to lint, can you fix it?
* can you add the references about the bedrock model and how to use it in the docs?

Thanks a lot!

gventuri:
Also, it would be cool, as we are implementing bedrock, to make it more generic, so that every model from bedrock can be used! What do you think? @goriri

goriri:
Hi @gventuri Sure. I can add some doc in llms.md for the Bedrock Claude 3. The coding capability is not as good as Claude 3 for other models in Bedrock, so the code generation/execution might fail. I'll try to add more Claude 3 models or other capable models when they're available.

gventuri:
@goriri thanks a lot, merging :D

PR #1068, opened 2024-03-27
fix(connectors): revert pandas to previous version temporarily

ArslanSaleem:


YarShev:
Hi @ArslanSaleem, what is the reason for this revert?

gventuri:
@YarShev the new pandas version is breaking all the SQL connectors. We need to figure out how to make it work with the connectors

YarShev:
@gventuri, ok, got it. Could you please keep me in tune with this, for instance, by tagging me in the issues that are going to be resolved on this matter.

PR #1075, opened 2024-03-29
docs: revert changes in pandas imports

mspronesti:
This PR reverts this [commit](https://github.com/Sinaptik-AI/pandas-ai/commit/331c7df29ed5b545a108c3cccd957bca3e0a3ac1).

**Rationale:** `pandasai.pandas` needs to be imported only within the package. The user is supposed to either import `pandas` or `modin.pandas` according to their desired engine, thus the docs and the examples shouldn't import `pandasai.pandas` but `pandas` (or `modin.pandas`).


gventuri:
@mspronesti this thing is still a little bit confusing to me. Took some time to look deeper in the implementation and now I think I got it. Thanks a lot for the PR, merging!

PR #1106, opened 2024-04-11
Upgrade Modin and pandas versions

### LINKED ISSUE #1085: Pandas 2.0 conflicts with the SQL connectors
### System Info

OS version: any
Python version: any
pandasai: any

### 🐛 Describe the bug

Our current setup is facing issues when attempting to upgrade to pandas versions beyond 2.0, as this update seems to interfere with the functionality of our SQL connectors. It's essential for us to delve into the root cause of this problem to identify a viable solution. This will enable us to successfully transition to a more recent version of pandas, as well as to the latest version of modin.

YarShev:
- [x] Closes #1085 (Replace xxxx with the GitHub issue number).
- [x] Tests passed if fixing a bug or adding a new feature.


YarShev:
Hey folks, I tried to upgrade Modin and pandas versions to latest once again as I didn't see any error in tests locally. Let's see how CI goes.

gventuri:
@YarShev IIRC, the issues were with the SnowFlake connector and probably others as well, but I'm not sure it is highlighted from the tests, as it might be mocked (which is not good, as there should be an error failing because of that, but I'm afraid we don't have a test for that yet).

YarShev:
@gventuri, how did you figure out that there is an issue with the Snowflake connector? Is there an open issue a user filed on this matter?

gventuri:
@ArslanSaleem initially reported it. Maybe he can elaborate further!

sfc-gh-mvashishtha:
> Maybe he can elaborate further!

Let's follow up on #1085.

PR #1109, opened 2024-04-14
fix: add df to environment for the good measure

oedokumaci:
This is to avoid cases where the generated code uses df, which gives the following error: NameError: name 'df' is not defined


gventuri:
Good addition, I would only conditionally passing it in case only one dataframe is passed, or we risk to pass the wrong one

PR #1111, opened 2024-04-15
fix: multi-round conversion with Bedrock Claude

goriri:


gventuri:
@goriri can you provide slightly more context about this fix? I don't understand exactly why we are overriding the previous message in the conversation. Does bedrock raise an error if there are 2 user messages in a row?

gventuri:
@goriri let me know, so we can merge :)

goriri:
@gventuri 
yes. it will complain when two user messages in a row (requiring alternating user/assistant messages). in a single message, the content can contain multiple text sections.

gventuri:
Thanks a lot for letting me know, merging! @goriri

Issue #1128, opened 2024-04-21
'No code found in the response' error - retry code gen with chatgpt


gDanzel:
### System Info

pandasai: 2.0.35
python: 3.11
OS: win11

### 🐛 Describe the bug

when error correction framework triggered, it's likely  the LLM  respondes without code delimiter '''' (in my case Chatgpt 3.5 turbo, like 3 out of 5 times),  and pandasai returned 'No code found in the response' error and breaks the pipe.  If the delimiter are enforced, would it be good to request the LLM for the delimiter of code in response?

```
import pandasai.pandas as pd
from pandasai import Agent
from pandasai.helpers import get_openai_callback
from pandasai.llm import OpenAI, GoogleGemini

from data.sample_dataframe import dataframe

llm = OpenAI()
agent = Agent([pd.DataFrame(dataframe)], config={"llm": llm, "enforce_privacy": True, "verbose": True, "enable_cache": False,})
with get_openai_callback() as cb:
    response = agent.chat("Calculate average, mean, median of DGP and hapineess_index.")
    print(response)
    print(cb)
```

The trace
```
C:\Users\Danzel\PycharmProjects\Tools\venv\Scripts\python.exe C:\Users\Danzel\PycharmProjects\Tools\yunai\privacyenforced.py 
2024-04-22 01:02:15 [INFO] Question: Calculate average, mean, median of DGP and hapineess_index.
2024-04-22 01:02:16 [INFO] Running PandasAI with openai LLM...
2024-04-22 01:02:16 [INFO] Prompt ID: 5b90df7a-261a-4442-a096-3a52c7387557
2024-04-22 01:02:16 [INFO] Executing Pipeline: GenerateChatPipeline
2024-04-22 01:02:16 [INFO] Executing Step 0: ValidatePipelineInput
2024-04-22 01:02:16 [INFO] Executing Step 1: CacheLookup
2024-04-22 01:02:16 [INFO] Executing Step 2: PromptGeneration
2024-04-22 01:02:19 [INFO] Using prompt: <dataframe>
dfs[0]:10x3
country,gdp,happiness_index
Germany,2518917471,7.23
China,5877894258,5.87
Spain,3873144538,6.38
</dataframe>




Update this initial code:
```python
# TODO: import the required dependencies
import pandas as pd

# Write code here

# Declare result var: 
[...]

PR #1140, opened 2024-05-01
fix:Introduce `_is_python_code` check into `_extract_code`

adamingas:
* Add tests to cover more cases for `_extract_code` and `_is_python_code`

- [X] Addresses #1128 
- [X] Tests added and passed if fixing a bug or adding a new feature.

Motivation:
Using a dataframe with a large number of columns, the `LLM()._execute_code` method would fail when asking how many rows there are in the dataframe. The failure would be because the LLM wasn't using triple back ticks to surround the code.

If the LLM produces correct python code, but for some reason does not
preface it with \```, the `LLM()._extract_code` method fails with
NoCodeFoundError.

In this change we first check if the separator is included in the response and we take only the part that is in-between. Then we clean the code from `python/py` statements.
Then we check if it's valid code. This check should pass otherwise executing the code will fail, and it's better to raise an error earlier than later.

With this change, if the LLM produced perfect code without the \``` separator, it would pass and return the code, whereas previously it would fail.

### DIFF
--- a/pandasai/llm/base.py
+++ b/pandasai/llm/base.py
@@ -69,7 +69,7 @@ class LLM:
     def _polish_code(self, code: str) -> str:
         """
         Polish the code by removing the leading "python" or "py",  \
-        removing the imports and removing trailing spaces and new lines.
+        removing surrounding '`' characters  and removing trailing spaces and new lines.
 
         Args:
             code (str): A string of Python code.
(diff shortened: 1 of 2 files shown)


tushar-31093:
Guys, this is a genuine concern, could you confirm if this fix might solve this problem?

adamingas:
@gventuri Can someone take a look at the PR? Happy to change stuff if need be.

gventuri:
@adamingas merged, thanks a lot for the PR :D

PR #1258, opened 2024-06-28
Fix issue #1257 empty dataframe

### LINKED ISSUE #1257: Empty dataframe is generated in code execution stage, which result in empty chart
### System Info

The current version of pandasai being used: 2.2.7

### 🐛 Describe the bug

In some case LLM will generate code to retrieve different subset of a same dataframe. In the example below it's generating the "first_purchases" dataframe as well as "repurchases" dataframe. Then these two dataframes are merged for further analysis, which makes perfect sense:

```
# Assuming dfs[0] is the DataFrame we need to work with
df = dfs[0]

# Filter out first-time purchases and repurchases
first_purchases = df[df['buyer_purchase_sequence'] == 1][['buyer_uid', 'product_type']]
repurchases = df[df['buyer_purchase_sequence'] > 1][['buyer_uid', 'product_type']]

# Merge first purchases with repurchases to track the product type of the first purchase against repurchases
merged_df = pd.merge(repurchases, first_purchases, on='buyer_uid', suffixes=('_repurchase', '_first_purchase'))
```

In the `code_execution` phase, two filters are extracted: "buyer_purchase_sequence == 1" (for "first_purchases" dataframe) and "buyer_purchase_sequence > 1" (for "repurchases" dataframe).

However in `pandasai/connectors/sql.py`, this will concatenate a where clause: **`where buyer_purchase_sequence == 1 AND buyer_purchase_sequence > 1`**. This will result an empty dataframe in dfs, and eventually an empty chart.

I understand the design of filtering dataframe is for better system performance. But considering all possible cases, the "where clause" needs to use `OR` instead of `AND` to ensure all data gets loaded in dfs.

xuyuanme:
- [x] Closes #1257.

As described in #1257: considering all possible cases, in `pandasai/connectors/sql.py` the "where clause" needs to use OR instead of AND to ensure all data gets loaded in dfs.

### DIFF
--- a/pandasai/connectors/sql.py
+++ b/pandasai/connectors/sql.py
@@ -173,7 +173,7 @@ class SQLConnector(BaseConnector):
                         query_params[f"value_{i}"] = value
 
             if condition_strings:
-                where_clause = " AND ".join(condition_strings)
+                where_clause = " OR ".join(condition_strings)
                 base_query = base_query.where(
                     text(where_clause).bindparams(**query_params)
                 )


gventuri:
@xuyuanme it is indeed expected to be a "AND". We need to come up with a new syntax for the "OR" in my opinion, otherwise we risk some breaking changes

xuyuanme:
Well received! I think we do need an "OR" syntax, otherwise in case described in #1257, it results a bug of empty data frame and chart.

xuyuanme:
Thanks @gventuri . I think the syntax itself should work. One problem is, the conditions (in #1257) are generated from `self._additional_filters`, instead of `self.config.where`.

`self._additional_filters` are created by function `_extract_filters()` in `code_execution.py`, which returns:
```
        Returns:
            dict: The dictionary containing all filters parsed from
                the passed code. The dictionary has the following structure:
                {
                    "<df_number>": [
                        ("<left_operand>", "<operator>", "<right_operand>")
                    ]
                }
```

This code does not read user config to generate the sql. So it does not have a place to config whether it's "AND" or "OR". Any thought on how this part should be handled?

gventuri:
@xuyuanme I recommend in that case to pass in the configs "direct_sql": True.
In this way, PandasAI will generate directly the SQL query with the more appropriate OR!

xuyuanme:
Ok let me do more tests against it. Thanks for the guide!

xuyuanme:
Close this PR.

Issue #1265, opened 2024-07-02
Unable To Generate Schema:  InvalidLLMOutputType (Response validation failed!)


mayankagarwal19911:
### System Info

pandasai 2.2.8
Python 3.9.6

### Sample code
`df = pd.DataFrame(columns=["column1"], data=[["value1"]])`
`df.head()`
`semantic_agent = SemanticAgent(dfs=df)`
`print(semantic_agent._schema)
`

### Error
`in BaseAgent.call_llm_with_prompt(self, prompt)`
`   return result`
    `else:`
          `raise InvalidLLMOutputType("Response validation failed!")`
  ` except Exception:`
       `if (`
            `not self.context.config.use_error_correction_framework`
             `or retry_count >= self.context.config.max_retries - 1`
     ` ):`

pratik9409:
#1265 Will like to solve this bug

gventuri:
@pratik9409 sure, thanks a lot for the availability, I've assigned it to you! :D

pratik9409:
# Testing with a simple DataFrame
df = pd.DataFrame(columns=["Empdata"], data=[[1], [2]])

df.head()


try:
    #  Create an instance of the SemanticAgent with the provided dataframe
    semantic_agent = SemanticAgent(dfs=df)
    #  Print the generated schema
    print(semantic_agent._schema)
except InvalidLLMOutputType as e:
    #  If the LLM fails to generate a valid schema, catch the InvalidLLMOutputType exception
    print(f"Error: {e}")  # Print the error message
    print("Using fallback schema...")  # Inform the user that a fallback schema will be used


![semanticouput](https://github.com/Sinaptik-AI/pandas-ai/assets/67755812/25d38f47-e61f-44d2-9520-92d49fa44c3b)

PR #1276, opened 2024-07-11
fixed bug #1265

pratik9409:
#1265 

# Testing with a simple DataFrame
df = pd.DataFrame(columns=["Empdata"], data=[[1], [2]])

df.head()

try:
    #  Create an instance of the SemanticAgent with the provided dataframe
    semantic_agent = SemanticAgent(dfs=df)
    #  Print the generated schema
    print(semantic_agent._schema)
except InvalidLLMOutputType as e:
    #  If the LLM fails to generate a valid schema, catch the InvalidLLMOutputType exception
    print(f"Error: {e}")  # Print the error message
    print("Using fallback schema...")  # Inform the user that a fallback schema will be used

semanticouput


pratik9409:
[gventuri](https://github.com/gventuri) Please check the code

examples/semantic_agent.py
```diff
@@ -57,8 +64,37 @@
...
+
+# Testing with a simple DataFrame
+df = pd.DataFrame(columns=["Empdata"], data=[[1], [2]])
+
+df.head()
+
+
+try:
+    #  Create an instance of the SemanticAgent with the provided dataframe
+    semantic_agent = SemanticAgent(dfs=df)
```
ArslanSaleem, line 88: This example need to reverted because it's not the valid use SemanticAgent we can definitely add comment about the custom schema if some one want to use different one but that too should be in documentation

pyproject.toml
```diff
@@ -53,7 +53,7 @@ ibm-watsonx-ai = { version = "^0.2.3", optional = true,  markers = "python_versi
 cx-Oracle = { version = "^8.3.0", optional = true }
 pinecone-client = { version = "^4.1.0", optional = true, markers = "python_version >= '3.10'"}
 sqlglot = {extras = ["rs"], version = "^25.0.3"}
-
+python-multipart
```
ArslanSaleem, line 56: I don't think we need python-multipart should be removed

server/.env.environment
```diff
@@ -0,0 +1,10 @@
+# Database
```
ArslanSaleem, line 1: Remove envs as .env.example already pushed and also it has your private pandasai_api_key

server/pyproject.toml
```diff
@@ -28,9 +28,6 @@ pyyaml = "^6.0.1"
 passlib = "^1.7.4"
 bcrypt = "3.1.7"
 python-dotenv = "^1.0.1"
-psycopg2 = "^2.9.9"
```
ArslanSaleem, line 31: Why did you removed them from here?

server/Dockerfile
```diff
@@ -28,6 +30,9 @@ RUN poetry lock --no-update
 # Install Python dependencies
 RUN poetry install --no-root
 
+# Install Python dependencies
+RUN pip install python-multipart
```
ArslanSaleem, line 34: Why add it manually we can put it in pyproject.yml for server

server/core/server.py
```diff
@@ -42,7 +41,6 @@ def on_auth_error(request: Request, exc: Exception):
 
 def init_routers(app_: FastAPI) -> None:
     app_.include_router(router)
-    app_.include_router(ee_router)
```
ArslanSaleem, line 45: What's the reason to remove ee_router?

server/.env.environment
pratik9409: Removed the env file

server/pyproject.toml
pratik9409: Bymistake it would have been removed, added back

server/core/server.py
pratik9409: Added back, must be by mistakenly got removed.

pratik9409:
# Testing with a sample DataFrame
df = pd.DataFrame(columns=["Empdata"], data=[[1], [2]])

df.head()


semantic_agent = SemanticAgent(dfs=df)
print(semantic_agent._schema)
sinaptik_output

gventuri:
Hi @pratik9409 thanks a lot for the PR, but I'm struggling a little bit to understand the scope of it. 
It seems to be changing many things and as a result, so many files changed in the frontend, backend and library. Can you try to clean it up and add a more detailed description of what's the scope?

PR #1341, opened 2024-08-28
fix bug in malicious code scan

### LINKED ISSUE #1340: _is_malicious_code doesn't look for whole word
### System Info

Pandas AI version: 2.2.14
Python Version: 3.10.0

### 🐛 Describe the bug

I was trying to run a query where I had mentioned OSE, and I got the error
```
"Code shouldn't use 'os', 'io' or 'chr', 'b64decode' functions as this could lead to malicious code execution."
```

So I went to [code_cleaning.py](https://github.com/Sinaptik-AI/pandas-ai/blob/main/pandasai/pipelines/chat/code_cleaning.py) and saw this line of code 
```python     
return any(module in code for module in dangerous_modules)
```
This was looking for just presence of the words instead of the whole word,
So it returned true for OSE and it contains os

shoebham:
- [x] Closes #1340 (Replace xxxx with the GitHub issue number).
- [ ] Tests added and passed if fixing a bug or adding a new feature.

### DIFF
--- a/pandasai/pipelines/chat/code_cleaning.py
+++ b/pandasai/pipelines/chat/code_cleaning.py
@@ -176,7 +176,7 @@ Code running:
             "(chr",
             "b64decode",
         ]
-        return any(module in code for module in dangerous_modules)
+        return any(re.search(r'\b'+re.escape(module)+r'\b',code) for module in dangerous_modules)
 
     def _is_jailbreak(self, node: ast.stmt) -> bool:
         """


ArslanSaleem:
@shoebham Thanks for the identifying that but this change won't be able to handle scenarios like "import pandas.io" it can still use os and io library from whitelisted libraries to attack.

ArslanSaleem:
@shoebham also let's add some test cases as well.

PR #1586, opened 2025-02-04
feature(Views): adding view for local sources

scaliseraoul:

### DIFF
--- a/pandasai/agent/base.py
+++ b/pandasai/agent/base.py
@@ -28,6 +28,7 @@ from pandasai.vectorstores.vectorstore import VectorStore
 
 from ..config import Config
 from ..constants import LOCAL_SOURCE_TYPES
+from ..data_loader.duck_db_connection_manager import DuckDBConnectionManager
 from .state import AgentState
 
 
@@ -119,16 +120,10 @@ class Agent:
 
     def _execute_local_sql_query(self, query: str) -> pd.DataFrame:
         try:
-            # Use a context manager to ensure the connection is closed
-            with duckdb.connect() as con:
-                # Register all DataFrames in the state
-                for df in self._state.dfs:
-                    con.register(df.schema.name, df)
-
-                # Execute the query and fetch the result as a pandas DataFrame
-                result = con.sql(query).df()
-
-            return result
+            db_manager = DuckDBConnectionManager()
+            for df in self._state.dfs:
+                db_manager.register(df.schema.name, df)
+            return db_manager.sql(query).df()
         except duckdb.Error as e:
             raise RuntimeError(f"SQL execution failed: {e}") from e
 
--- a/pandasai/agent/base.py
+++ b/pandasai/agent/base.py
@@ -145,13 +145,13 @@ class Agent:
         if not self._state.dfs:
             raise ValueError("No DataFrames available to register for query execution.")
 
-        if (
-            self._state.dfs[0].schema.source
-            and self._state.dfs[0].schema.source.type in LOCAL_SOURCE_TYPES
-        ):
+        df0 = self._state.dfs[0]
+        source = df0.schema.source or None
+
+        if source and source.type in LOCAL_SOURCE_TYPES:
             return self._execute_local_sql_query(query)
         else:
-            return self._state.dfs[0].execute_sql_query(query)
+            return df0.execute_sql_query(query)
 
     def execute_with_retries(self, code: str) -> Any:
         """Execute the code with retry logic."""
--- a/pandasai/agent/base.py
+++ b/pandasai/agent/base.py
@@ -123,7 +123,7 @@ class Agent:
             with duckdb.connect() as con:
                 # Register all DataFrames in the state
                 for df in self._state.dfs:
-                    con.register(df.schema.source.table, df)
+                    con.register(df.schema.name, df)
 
                 # Execute the query and fetch the result as a pandas DataFrame
                 result = con.sql(query).df()
@@ -145,7 +145,10 @@ class Agent:
         if not self._state.dfs:
             raise ValueError("No DataFrames available to register for query execution.")
 
-        if self._state.dfs[0].schema.source.type in LOCAL_SOURCE_TYPES:
+        if (
+            self._state.dfs[0].schema.source
+            and self._state.dfs[0].schema.source.type in LOCAL_SOURCE_TYPES
+        ):
             return self._execute_local_sql_query(query)
         else:
             return self._state.dfs[0].execute_sql_query(query)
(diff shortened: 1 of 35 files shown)


pandasai/dataframe/base.py
```diff
@@ -260,11 +260,10 @@ def get_default_schema(cls, dataframe: DataFrame) -> SemanticLayerSchema:
         )
 
         return SemanticLayerSchema(
-            name=f"{dataframe._column_hash}",
```
gventuri, line 263: Is `_column_hash` still needed somewhere else?

pandasai/query_builders/base_query_builder.py
```diff
@@ -50,12 +41,19 @@ def _add_limit(self, n=None) -> str:
...
-        order_by = "RANDOM()" if source_type in {"sqlite", "postgres"} else "RAND()"
-        return f"{query} ORDER BY {order_by} LIMIT {n}"
+        return f"{query} LIMIT {n}"
 
     def get_row_count(self):
         return f"SELECT COUNT(*) {self._get_from_statement()}"
+
+    def _get_columns(self) -> str:
+        if self.schema.columns:
+            return ", ".join([col.name for col in self.schema.columns])
```
gventuri, line 54: Here we risk a SQL injection, don't we? It's true that is the user that provides the col names, but it's also true that if we don't sanitize, we are at risk. What's your take?

pandasai/query_builders/base_query_builder.py
```diff
@@ -50,12 +41,19 @@ def _add_limit(self, n=None) -> str:
...
+        else:
+            return "*"
+
+    def _get_from_statement(self):
+        return f"FROM {self.schema.name}"
```
gventuri, line 59: Same as above

pandasai/query_builders/base_query_builder.py
```diff
@@ -50,12 +41,19 @@ def _add_limit(self, n=None) -> str:
         return f" LIMIT {limit}" if limit else ""
 
     def get_head_query(self, n=5):
-        source_type = self.schema.source.type
         columns = self._get_columns()
         query = f"SELECT {columns} "
         query += self._get_from_statement()
-        order_by = "RANDOM()" if source_type in {"sqlite", "postgres"} else "RAND()"
-        return f"{query} ORDER BY {order_by} LIMIT {n}"
+        return f"{query} LIMIT {n}"
```
gventuri, line 47: Check comment below

gventuri (changes requested):
All good overall, only left a couple of changes!

pandasai/dataframe/base.py
gventuri: Ok thanks for clarifying, just wanted to avoid to have leftovers :)

pandasai/query_builders/view_query_builder.py
```diff
@@ -74,13 +59,7 @@ def _get_from_statement(self):
             loader = self.schema_dependencies_dict[to_datasets]
             subquery, dataset_name = self._get_sub_query_from_loader(loader)
             query += f"\nJOIN ( {subquery} ) AS {dataset_name}\n"
-            query += f"ON {relation.from_} = {relation.to}"
+            query += f"ON {sanitize_relation_name(relation.from_)} = {sanitize_relation_name(relation.to)}"
```
ArslanSaleem, line 62: I think we should protect all query builders from malicious attacks either we come up with some solid sanitization function or using parametrized queries.

pandasai/query_builders/base_query_builder.py
ArslanSaleem: Yes, it's sql injectable we either use parametrized queries or sanitize at the level of pydantic schema validation.

ArslanSaleem (changes requested):
Overall looks very good. Found two issues apart for sql injection.
1. In pull method it uses `dataset_loader.load(self.path) ` where load method doesn't take any argument.
2. In case of sql connector it is failing in `code_cleaning._clean_sql_query` where allowed table is passed as schema.name which doesn't exists in the database.

scaliseraoul:
Hi @ArslanSaleem , good catch for n1, could you iterate on n2,
"In case of SQL connector, it is failing in code_cleaning._clean_sql_query where allowed table is passed as schema.name which doesn't exist in the database." 

cause schema.name it's the dataset name, and it's always present

pandasai/agent/base.py
```diff
@@ -117,18 +128,20 @@ def execute_code(self, code: str) -> dict:
...
-                    con.register(df.schema.name, df)
+    @staticmethod
+    def _parse_correct_table_name(query: str, dfs: List[VirtualDataFrame]) -> str:
+        table_mapping = {
+            df.schema.name: df.query_builder._get_table_expression() for df in dfs
+        }
 
-                # Execute the query and fetch the result as a pandas DataFrame
-                result = con.sql(query).df()
+        return SQLParser.replace_table_and_column_names(query, table_mapping)
```
gventuri, line 137: Great use of sqlglot, didn't think about it

pandasai/query_builders/base_query_builder.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            return ["*"]
+
+    def _get_table_expression(self) -> str:
+        return self.schema.name
+
+    @staticmethod
+    def check_compatible_sources(sources: List[Source]) -> bool:
+        base_source = sources[0]
+        return all(base_source.is_compatible_source(source) for source in sources[1:])
```
gventuri, line 47: Can we add tests to make sure that no SQL injection can be done with the following fields?
- self.schema.name
- self.schema.columns

gventuri: I'm not sure sqlglot does sanitization by default

gventuri: Well done, very elegant architecture tho!

tests/unit_tests/data_loader/test_sql_loader.py
```diff
@@ -46,48 +43,12 @@ def test_load_mysql_source(self, mysql_schema):
...
             mock_execute_query.assert_called_once_with(
-                "SELECT email, first_name, timestamp FROM users ORDER BY RAND() LIMIT 5"
-            )
-
-            # Test executing a custom query
-            custom_query = "SELECT email FROM users WHERE first_name = 'John'"
-            result.execute_sql_query(custom_query)
-            mock_execute_query.assert_called_with(custom_query)
-
-    def test_mysql_schema(self, mysql_schema):
```
gventuri, line 57: Why removing the `test_mysql_schema`?

gventuri (changes requested):
Great improvements, only a couple of leftovers

tests/unit_tests/data_loader/test_sql_loader.py
scaliseraoul: it's a duplicate of test_load_mysql_source

PR #1894, opened 2026-06-08
Reject NaN/inf numeric results instead of returning them as a valid answer

aaravanmay:
`ResponseParser._validate_response` accepts a `number` result via `isinstance(result["value"], (int, float, np.int64))`. But `NaN` and `inf` are floats, so they pass that check, get wrapped in a `NumberResponse`, and are returned as the answer — with no error.

This is the silent-wrong case: when generated code aggregates over an **empty** result — e.g. `df["sales"].mean()` after a filter that matched zero rows — pandas returns `nan`. The user (and any downstream charting code) then receives `nan` as a confident numeric answer with no indication anything went wrong.

The fix adds a finite check after the existing numeric `isinstance` check:

```python
if isinstance(result["value"], float) and not np.isfinite(result["value"]):
    raise InvalidOutputValueMismatch(
        "Invalid output: Numeric result is NaN or infinite (likely an aggregation over empty data)."
    )
```

`np` is already imported; `np.isfinite` covers both NaN and inf, and `np.float64` subclasses `float` so it's covered too. Booleans/ints are unaffected.

Includes a regression test (no LLM call) asserting NaN and inf both raise `InvalidOutputValueMismatch` while a normal number still parses.

Found while building a fault-injection tester for agent tools.


aaravanmay:
This NaN-as-a-valid-answer bug is the exact failure class I built faultline (open source) to catch: a data agent confidently returning a number that's wrong or not-a-number, with no error raised. The numeric-finite invariant that catches it is ~3 lines, and it runs deterministically in CI — no LLM judge. I've drafted a small suite for pandas-ai covering this class (NaN/inf results, silently truncated dataframes). Want me to open it as a separate PR with a non-blocking GitHub Action so you can watch what it flags for a week before deciding? If it's noisy, close it — no hard feelings.

